PR #2, opened 2022-06-22
feat: MVP of ABI generation

itegulov:
This iteration of cargo-near follows the new approach with dynamic linking to a shared library that I presented before.

~A couple of places presume ELF (filtering by `.so` suffix, stripping `lib` prefix from the shared library), but I think with minimal tweaks the same will work for MachO and PE.~ UPD: made some tweaks, works on Linux & macOS (M1, but should not matter I think). I don't have easy access to a Windows machine so that would have to be done a bit later.

`prettyprint` and `syn` dependencies could be removed, but I thought generating readable code would be nice (especially since I am logging it for visibility).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "cargo-near"
+version = "0.1.0"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/near/cargo-near"
+description = """
+Cargo extension for building Rust smart contracts on NEAR.
+"""
+
+[dependencies]
+anyhow = "1.0"
+cargo_metadata = "0.14"
+clap = { version = "3.2", features = ["derive", "env"] }
+colored = "2.0"
+env_logger = "0.9"
+near-sdk = { git = "https://github.com/near/near-sdk-rs.git", branch = "daniyar/abi-macro", features = ["abi"] }
+log = "0.4"
+prettyplease = "0.1"
+tempfile = "3.3"
+toml = "0.5"
+serde_json = "1.0"
+symbolic-debuginfo = "8.8"
+syn = "1.0"
+quote = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ cargo_metadata = "0.14"
 clap = { version = "3.2", features = ["derive", "env"] }
 colored = "2.0"
 env_logger = "0.9"
-near-sdk = { git = "https://github.com/near/near-sdk-rs.git", ref = "9ab6df8ed35e67c247299059d8277c2170a0a91b", features = ["abi"] }
+near-sdk = { git = "https://github.com/near/near-sdk-rs.git", rev = "9ab6df8ed35e67c247299059d8277c2170a0a91b", features = ["abi"] }
 log = "0.4"
 prettyplease = "0.1"
 toml = "0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ cargo_metadata = "0.14"
 clap = { version = "3.2", features = ["derive", "env"] }
 colored = "2.0"
 env_logger = "0.9"
-near-sdk = { git = "https://github.com/near/near-sdk-rs.git", branch = "daniyar/abi-macro", features = ["abi"] }
+near-sdk = { git = "https://github.com/near/near-sdk-rs.git", ref = "9ab6df8ed35e67c247299059d8277c2170a0a91b", features = ["abi"] }
 log = "0.4"
 prettyplease = "0.1"
 tempfile = "3.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,6 @@ env_logger = "0.9"
 near-sdk = { git = "https://github.com/near/near-sdk-rs.git", ref = "9ab6df8ed35e67c247299059d8277c2170a0a91b", features = ["abi"] }
 log = "0.4"
 prettyplease = "0.1"
-tempfile = "3.3"
 toml = "0.5"
 serde_json = "1.0"
 symbolic-debuginfo = "8.8"
--- a/README.md
+++ b/README.md
@@ -1,2 +1,12 @@
 # cargo-near
 Cargo extension for building Rust smart contracts on NEAR
+
+To install:
+```
+cargo install --path .
+```
+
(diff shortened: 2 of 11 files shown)


itegulov:
@miraclx assigned you to the PR review since you might be interested, but no pressure ofc

miraclx (approved):
Nice work, lgtm!

README.md
```diff
@@ -1,2 +1,12 @@
 # cargo-near
 Cargo extension for building Rust smart contracts on NEAR
+
+To install:
+```
+cargo install --path .
+```
+
+To generate ABI for a contract (while standing in the directory containing contract's Cargo.toml):
```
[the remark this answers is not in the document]
itegulov: Nah not a rule, clarified README

src/abi/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+    let main_rs = generation::generate_main_rs(&dylib_path)?;
+    fs::write(tmp_dir.path().join("main.rs"), main_rs)?;
+
+    let target_dir_arg = format!(
+        "--target-dir={}",
+        crate_metadata.target_directory.to_string_lossy()
+    );
+    let stdout = util::invoke_cargo(
+        "run",
+        &["--package", "near-abi-gen", "--release", &target_dir_arg],
```
miraclx, line 44: This caches the build artifacts in the target folder of the contract, and not the temp directory. Is this intended?

src/abi/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+/// ABI generation result.
+pub(crate) struct AbiResult {
+    /// Path to the resulting ABI file.
+    pub path: PathBuf,
+}
+
+pub(crate) fn execute(manifest_path: &CargoManifestPath) -> anyhow::Result<AbiResult> {
+    let tmp_dir = tempfile::Builder::new()
+        .prefix("cargo-contract_")
+        .tempdir()?;
```
miraclx, line 20: What do you think about persisting the code generated ABI builder source? I, for one, would like to be able to inspect it.

Especially since we're reusing the same target directory of the contract to host the build artifacts of the code generated ABI builder. So we'll have something of the format:

```console
hello-world
├── Cargo.toml
├── src
│   └── lib.rs
└── target
    ├── near
    │   └── hello-world
    │       ├── build.rs
    │       ├── Cargo.toml
    │       ├── src
    │       │   └── lib.rs
    │       └── target
    │           ├── abi.json
    │           └── release
    │               └── hello-world
    └── release
        └── libhello_world.so
```

miraclx (commented):
Just a few more things.

src/abi/generation.rs
```diff
@@ -0,0 +1,149 @@
...
+    let original_cargo_toml = fs::read_to_string(&manifest_path.path)?;
+    let original_cargo_toml: toml::value::Table = toml::from_str(&original_cargo_toml)?;
+
+    let mut near_sdk = original_cargo_toml
+        .get("dependencies")
+        .ok_or_else(|| anyhow::anyhow!("[dependencies] section not found"))?
+        .get("near-sdk")
+        .ok_or_else(|| anyhow::anyhow!("near-sdk dependency not found"))?
+        .as_table()
+        .ok_or_else(|| anyhow::anyhow!("near-sdk dependency should be a table"))?
```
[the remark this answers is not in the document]
itegulov: My assumption is that `abi` is going to stay a feature in `near-sdk` (that might become enabled by default in the future).

@austinabell do you agree with my understanding?

src/abi/mod.rs
itegulov: Yes, I assumed this way I would be able to leverage incremental compilation if needed. Is there an argument for keeping this in the tmp directory? I guess the build artifact sizes can be pretty significant.

src/abi/mod.rs
itegulov: Hmmm, yeah I don't see any immediate downsides to this and this is definitely good for transparency and debugging. I'll try this out.

src/abi/mod.rs
miraclx: Yeah, that makes sense. Just often used to having source and build artifacts together. Maybe we should consider keeping the builder source together with its build artifacts instead of a temp dir?

austinabell: yeah, I would intuitively think that keeping as is and not using a tmp directory for this would be ideal. Unsure how much would be saved by the incremental compilation, but I don't see the benefit in using tmp directory here

src/abi/generation.rs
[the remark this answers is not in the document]
austinabell: I agree, but also `abi` might be a default feature in the future, so this wouldn't have to be specifically defined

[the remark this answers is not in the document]
miraclx: yeah, so when it's not specifically defined, then the entry won't be a table. meaning this would force users to do this:

```toml
near-sdk = { version = "5.0.0" }
```

src/abi/mod.rs
itegulov: Done. This makes much more sense, thanks for the suggestions.

> Unsure how much would be saved by the incremental compilation

It is quite considerable since the generated `near-abi-gen` has to depend on `near-sdk` with all of its dependencies. So always losing the target directory does not seem like a good way to go about this.

Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+Cargo extension for building Rust smart contracts on NEAR.
+"""
+
+[dependencies]
+anyhow = "1.0"
+cargo_metadata = "0.14"
+clap = { version = "3.2", features = ["derive", "env"] }
+colored = "2.0"
+env_logger = "0.9"
+near-sdk = { git = "https://github.com/near/near-sdk-rs.git", branch = "daniyar/abi-macro", features = ["abi"] }
```
austinabell, line 18: noting that this should wait until https://github.com/near/near-sdk-rs/pull/831 comes in and then pulling a commit rather than a branch so this doesn't break when the branch is deleted

austinabell (approved):

Cargo.toml
itegulov: Yeah for sure

itegulov: Done

austinabell (approved):
lgtm, unsure if the windows things are fixed, but that can always be fixed later.

itegulov:
Windows things are unresolved unfortunately, created an issue here: https://github.com/near/cargo-near/issues/4

PR #3, opened 2022-06-28
feat: set up integration tests for ABI generation flow

itegulov:
Some sanity check integration tests. They take a little while to run but test pretty closely to what would be happening in a real scenario. LMK if there is a better way to do this.

Will set up Github Actions workflows in a follow up PR.


integration-tests/tests/tests/callbacks.rs
```diff
@@ -0,0 +1,174 @@
...
+        Some(AbiType {
+            type_schema: u32_schema,
+            serialization_type: AbiSerializationType::Json
+        })
+    );
+
+    Ok(())
+}
+
+// FIXME: test does not work because callback_vec type is not unwrapped from `Vec<_>`.
```
austinabell, line 110: updating the sdk commit and fixing this before bringing in or?

integration-tests/tests/tests/modifiers.rs
```diff
@@ -0,0 +1,134 @@
...
+fn test_non_init_function() -> anyhow::Result<()> {
+    let abi_root = generate_abi_fn! {
+        pub fn add(&self, a: u32, b: u32) -> u32 {
+            a + b
+        }
+    };
+
+    assert_eq!(abi_root.abi.functions.len(), 1);
+    let function = &abi_root.abi.functions[0];
+    assert!(!function.is_init);
```
austinabell, line 48: why not just add this check in `test_view_function` to avoid duplication?

integration-tests/tests/tests/modifiers.rs
```diff
@@ -0,0 +1,134 @@
...
+    let abi_root = generate_abi_fn! {
+        #[init]
+        pub fn add(a: u32, b: u32) -> Self {
+            Contract {}
+        }
+    };
+
+    assert_eq!(abi_root.abi.functions.len(), 1);
+    let function = &abi_root.abi.functions[0];
+    assert!(function.is_init);
```
austinabell, line 65: can you also add an assert that makes sure there is no return value in the ABI?

integration-tests/tests/tests/modifiers.rs
```diff
@@ -0,0 +1,134 @@
...
+fn test_non_payable_function() -> anyhow::Result<()> {
+    let abi_root = generate_abi_fn! {
+        pub fn add(&mut self, a: u32, b: u32) -> u32 {
+            a + b
+        }
+    };
+
+    assert_eq!(abi_root.abi.functions.len(), 1);
+    let function = &abi_root.abi.functions[0];
+    assert!(!function.is_payable);
```
austinabell, line 81: same as before, can just add this assert to the previous test

austinabell (commented):
I really like these tests! This was very creative

integration-tests/tests/tests/callbacks.rs
itegulov: Yeah since they are already approved I will just wait until the PRs are in

integration-tests/tests/tests/modifiers.rs
itegulov: Felt like splitting the tests makes each of the things I am trying to test stand out. But given that they are all very small and each test increases the run time quite a bit I agree that it makes sense to collapse some of them together

austinabell (approved):

Issue #6, opened 2022-07-13
No error on `cargo near abi` when abi feature on SDK contract is not set


austinabell:
A little bit of a foot gun where a basically empty ABI file will be generated in this case

PR #9, opened 2022-07-18
feat: throw an error when "abi" feature is not enabled

itegulov:
Fixes https://github.com/near/cargo-near/issues/6

### DIFF
--- a/cargo-near/src/abi/generation.rs
+++ b/cargo-near/src/abi/generation.rs
@@ -17,6 +17,15 @@ pub(crate) fn generate_toml(manifest_path: &CargoManifestPath) -> anyhow::Result
         .ok_or_else(|| anyhow::anyhow!("near-sdk dependency should be a table"))?
         .clone();
 
+    if !near_sdk
+        .get("features")
+        .and_then(|features| features.as_array())
+        .map(|features| features.contains(&value::Value::String("abi".to_string())))
+        .unwrap_or(false)
+    {
+        anyhow::bail!("Unable to generate ABI: NEAR SDK \"abi\" feature is not enabled")
+    }
+
     let cargo_toml = include_str!("../../templates/_Cargo.toml");
     let mut cargo_toml: toml::value::Table = toml::from_str(cargo_toml)?;
     let deps = cargo_toml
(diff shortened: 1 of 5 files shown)


cargo-near/src/abi/generation.rs
```diff
@@ -17,6 +17,15 @@ pub(crate) fn generate_toml(manifest_path: &CargoManifestPath) -> anyhow::Result
         .ok_or_else(|| anyhow::anyhow!("near-sdk dependency should be a table"))?
         .clone();
 
+    if !near_sdk
+        .get("features")
+        .and_then(|features| features.as_array())
+        .map(|features| features.contains(&value::Value::String("abi".to_string())))
+        .unwrap_or(false)
```
austinabell, line 24: will this fail if we ever make abi a default feat of the SDK? Assuming this is just checking the toml syntactically?

austinabell (approved):

itegulov:
> question before we merge this, do you think it's worth automatically setting this feature when using cargo near on the generated toml file?

I am already doing that for the generated toml file, but `abi` should also be set for the original crate as I need to compile a dylib containing ABI symbols.

In theory, I could clone the directory into `/tmp` and set `abi` feature there, but that could mess up relative paths. Not sure if there is a cleaner solution to this...

cargo-near/src/abi/generation.rs
itegulov: Yes, this is just a syntactic check. I tried to find a cargo command that would tell me if a feature on one of the dependencies is enabled (even implicitly e.g. by `default`), but couldn't find one.

austinabell:
> I am already doing that for the generated toml file, but `abi` should also be set for the original crate as I need to compile a dylib containing ABI symbols.

If you're already modifying the toml file to include dylib, can you not add the "abi" feature to the near-sdk dependency, in theory? Doesn't have to happen in this PR, just curious how we can make this more seamless

itegulov:
> If you're already modifying the toml file to include dylib

No, I am not doing that. I guess I am kind of assuming that all contracts have to have crate-type set to cdylib like this: https://github.com/near/near-sdk-rs/blob/master/examples/non-fungible-token/nft/Cargo.toml#L8. Is this not a fair assumption?

PR #10, opened 2022-07-22
feat: optimize ABI generation time

### LINKED ISSUE #8: Remove requirement to compile SDK testing utils with ABI generation
This issue could just as easily be opened in the SDK repo, but since it's more relevant to this I'm opening it here.

Roughly the pattern that would be easiest would be:
- Add `testing` feature to SDK, which gates the nearcore dependencies behind it and makes it default (has benefits to SDK)
- Modify crate toml to use the same version of the SDK but with no default features and "abi" feature (there will also have to be parsing to include other features like "unstable" that might be included). This also would have the benefit of not requiring the developer to enable abi feature on their crate by default

Maybe this is janky, is there a better way?

itegulov:
Fixes #8 

This PR introduces three compilation time improvements:
1. Removes `nearcore` crate dependencies from the generated abigen crates (reduces compilation time from 22.5s to 10.5s on `abi` example in NEAR SDK).
2. Makes the generated abigen crates use `dev` profile with disabled optimizations and disabled LTO (reduces compilation town from 10.5s further to 9.5s)
3. Makes the original (user-written) crate use `dev` profile with disabled optimizations and disabled LTO through cargo envvars (reduces dylib compilation time from 25s to 21.5s).

### DIFF
--- a/cargo-near/src/abi/generation.rs
+++ b/cargo-near/src/abi/generation.rs
@@ -26,7 +26,6 @@ pub(crate) fn generate_toml(manifest_path: &CargoManifestPath) -> anyhow::Result
         .expect("[dependencies] is a table specified in the template");
 
     // Make near-sdk dependency not use default features to save on compilation time, but ensure `abi` is enabled
-    near_sdk.remove("default-features");
     near_sdk.remove("optional");
     near_sdk.insert("default-features".to_string(), value::Value::Boolean(false));
     near_sdk.insert(
--- a/cargo-near/src/abi/generation.rs
+++ b/cargo-near/src/abi/generation.rs
@@ -25,9 +25,10 @@ pub(crate) fn generate_toml(manifest_path: &CargoManifestPath) -> anyhow::Result
         .as_table_mut()
         .expect("[dependencies] is a table specified in the template");
 
-    // Make near-sdk dependency use default features
+    // Make near-sdk dependency not use default features to save on compilation time, but ensure `abi` is enabled
     near_sdk.remove("default-features");
     near_sdk.remove("optional");
+    near_sdk.insert("default-features".to_string(), value::Value::Boolean(false));
     near_sdk.insert(
         "features".to_string(),
         value::Value::Array(vec![value::Value::String("abi".to_string())]),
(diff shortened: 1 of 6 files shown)


cargo-near/src/abi/generation.rs
```diff
@@ -25,9 +25,10 @@ pub(crate) fn generate_toml(manifest_path: &CargoManifestPath) -> anyhow::Result
         .as_table_mut()
         .expect("[dependencies] is a table specified in the template");
 
-    // Make near-sdk dependency use default features
+    // Make near-sdk dependency not use default features to save on compilation time, but ensure `abi` is enabled
     near_sdk.remove("default-features");
     near_sdk.remove("optional");
+    near_sdk.insert("default-features".to_string(), value::Value::Boolean(false));
```
austinabell, line 31: This will be an issue when a crate uses a feature like `unstable`, no?

cargo-near/src/abi/generation.rs
```diff
@@ -25,9 +25,10 @@ pub(crate) fn generate_toml(manifest_path: &CargoManifestPath) -> anyhow::Result
         .as_table_mut()
         .expect("[dependencies] is a table specified in the template");
 
-    // Make near-sdk dependency use default features
+    // Make near-sdk dependency not use default features to save on compilation time, but ensure `abi` is enabled
     near_sdk.remove("default-features");
```
austinabell, line 29: can remove this line then, right?

austinabell (commented):
LGTM, aside from just the potentially breaking change with turning off default features

cargo-near/src/abi/generation.rs
itegulov: The Cargo.toml I am generating here will be used only for building and running this bit of code: https://github.com/near/cargo-near/blob/main/cargo-near/src/abi/generation.rs#L130-L140.

The reason why I am transforming near-sdk dependency in such a backwards way is that there are so many ways to specify a dependency version that I did not want to deal with (path to a local directory, git URI + branch, git URI + commit hash etc) and instead just deleting stuff that is not needed.

cargo-near/src/abi/generation.rs
itegulov: Indeed

austinabell (approved):

PR #15, opened 2022-07-29
feat: add cargo dependency tests

itegulov:
Trying to identify all the different scenarios how a developer could have a near-sdk dependency in their Cargo.toml. A few corner cases are not working, I feel like it would be worth it to add `patch` section support, not sure about `renamed` and platform-specific dependencies. Either way, will be handled in a follow-up PR.

Also, I had to remove common Cargo workspace for all integration tests since it was causing a lot of hard-to-overcome issues (e.g. patch section should be in the root Cargo.toml, local dependency clashing with one pulled from Github). The downside is that it increases the test build time. Btw, if you have run the integration tests on your machine at least once in the past, I recommend doing `cargo clean` before running them on this PR as the residual files from the previous approach can cause issues with this PR.


itegulov:
Welp, now Github Actions are failing because there is not enough disk space on our runner 🥴

austinabell:
> Welp, now Github Actions are failing because there is not enough disk space on our runner 🥴

Hmm, we should probably drop the cache for these small libraries. I'll do that now

austinabell (approved):

PR #22, opened 2022-08-11
update near-sdk

miraclx:
- https://github.com/near/near-sdk-rs/pull/876
- https://github.com/near/near-sdk-rs/pull/877
- https://github.com/near/near-sdk-rs/pull/875


itegulov (commented):
Can you also update Cargo.toml files under https://github.com/near/cargo-near/tree/main/integration-tests/templates and these two values here: https://github.com/near/cargo-near/blob/main/integration-tests/tests/cargo/mod.rs#L22 & https://github.com/near/cargo-near/blob/main/integration-tests/tests/cargo/mod.rs#L43

Sorry this wasn't documented properly :)

PR #26, opened 2022-08-18
CI: drop MSRV tests

miraclx:
Context: https://github.com/near/cargo-near/pull/24#issuecomment-1218753388

Inherits some refactoring from #24.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Cargo NEAR Checks And Tests
+name: CI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,61 +1,46 @@
...
-        toolchain: [stable, 1.56.0]
+
     steps:
       - name: Checkout Sources
         uses: actions/checkout@v2
-      - name: "Install ${{ matrix.toolchain }} toolchain"
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: ${{ matrix.toolchain }}
```
miraclx, line 23: I wonder if it's worth it to retain these. By default, the runner ships with an up-to-date rust version. - https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#rust-tools

But we might want to keep this to ensure it always on the latest stable. WDYT?

itegulov: I think it is fine to use Github runner's native rust toolchain. Seems like they are keeping it up-to-date.

itegulov (approved):

PR #30, opened 2022-08-18
cargo check msrv

miraclx:
Context: #24. Supersedes #26.

Ensures `cargo-near` remains compilable with the MSRV.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
 
       - name: Get MSRV
         run: |
-          RUST_MSRV="$(cat cargo-near/Cargo.toml | sed -n 's/rust-version\s*=\s*"\(.*\)"/\1/p')"
+          RUST_MSRV="$(cat cargo-near/Cargo.toml | sed -n 's/rust-version *= *"\(.*\)"/\1/p')"
           echo "RUST_MSRV=$RUST_MSRV" >> $GITHUB_ENV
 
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -6,6 +6,30 @@ on:
...
+      matrix:
+        platform: [ubuntu-latest, macos-latest]
+
+    steps:
+      - name: Checkout Sources
+        uses: actions/checkout@v2
+
+      - name: Get MSRV
+        run: |
+          RUST_MSRV="$(cat cargo-near/Cargo.toml | sed -n 's/rust-version *= *"\(.*\)"/\1/p')"
```
miraclx, line 21: Cut some corners here for cross-platform compatibility, seems fine.. Or should we do this?

```suggestion
          RUST_MSRV="$(cat cargo-near/Cargo.toml | sed -n 's/rust-version[[:space:]]*=[[:space:]]*"\(.*\)"/\1/p')"
```

austinabell: if it works on CI, I don't personally see a downside

austinabell (approved):

.github/workflows/test.yml
miraclx: > Edit: are you thinking about windows compatibility?

Oh, no.. I haven't tested that. It was macOS I was thinking of. I used `\s` first, and it didn't support that for some reason, so I used a hard space.

Issue #21, opened 2022-08-04
Implement build command to build contract and ABI together


austinabell:
Motivation is that it would be preferred if these are built together to ensure there is no mismatch of versions when using.

```
cargo near build
```

should build similar to `cargo build` except that it includes `"-C", "link-arg=-s"` rustflags to ensure those aren't left out and includes `--target wasm32-unknown-unknown`. Unsure if all cargo options should/can be forwarded easily, but at least `--release` needs to be included as configuration.

This command should output the following files in `$CARGO_TARGET_DIR/near/`:

- `<crate name>.json`: ABI of contract (logic exists already with `cargo near abi`)
- `<crate name>.wasm`: Built wasm file
- `<crate name>.<extension>`: Bundle of ABI and wasm in JSON file. Wasm should probably be included in base64 (nearcore RPC accepts this, so no transformation will be required). The motivation for this is some tooling might want both to avoid having to use two separate files.

What the bundle looks like may be opinionated, maybe it's the same schema as ABI except with an extra key for `code`, or it's a map of `{"abi": ..., "code":}`. The latter might be a bit redundant/confusing because ABI already has a key `abi`.

If it works out cleanly, this can be split into three separate changes:

- [x] Build command that outputs `.wasm`
- [x] Build and include ABI during the command
- [ ] Create bundle output of both

austinabell:
@miraclx, you were looking into this and related tasks, do you want to assign yourself to this?

miraclx:
Sure thing.

austinabell:
re-opening because there are still components relevant in this issue. exalate screwed up the formatting so it's unviewable now. I'll try to fix it later

PR #31, opened 2022-08-21
ABI embedding

miraclx:
Based off #23.

Introduces the `cargo near build` subcommand with the `--embed-abi` flag.

Building without the `--embed-abi` flag is effectively analogous to the convention of using `./build.sh` to build contracts.

With the flag however, `cargo near` inspects the contract, generating ABI that describes the contract and thereafter, embeds it into the contract. Which can be returned by calling the view method `__contract_abi`.

Some numbers (compiling the `abi` example in sdk):

- Without embedding ABI

  ```console
  $ cargo near build --out-dir res --release
  $ wc -c < res/abi.wasm
  97313
  ```

- With embedded ABI

  ```console
  $ cargo near build --out-dir res --release --embed-abi
  $ wc -c < res/abi.wasm
  99081
  ```

That's an increase of 1768 bytes (1.78% of the final binary).

Related: https://github.com/near/near-sdk-rs/pull/893


cargo-near/src/contract.rs
```diff
@@ -0,0 +1,93 @@
...
+            &cargo_args,
+            vec![("CARGO_NEAR_ABI_PATH", abi_file_path.to_str().unwrap())],
+            "wasm",
+        )?
+    } else {
+        util::compile_project(&crate_metadata.manifest_path, &cargo_args, vec![], "wasm")?
+    };
+
+    let mut out_path = wasm_artifact.path;
+    if let Some(out_dir) = args.out_dir.and_then(|o| o.canonicalize().ok()) {
```
itegulov, line 81: This is going to silently swallow a potential error during canonicalization. Another thing we should probably do during the initialization - canonicalize all user-input paths and assume they are such from that point onwards (maybe even create a small zeor-cost wrapper around `PathBuf` - `CanonicalPathBuf`). Doesn't have to happen in this PR, just sharing my thoughts.

cargo-near/src/main.rs
```diff
@@ -1,7 +1,8 @@
-use cargo_near::Opts;
 use clap::Parser;
 use colored::Colorize;
 
+use cargo_near::Opts;
```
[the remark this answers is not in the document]
miraclx: Makes sense, I'll fix that.

cargo-near/src/contract.rs
```diff
@@ -0,0 +1,93 @@
+use std::env;
+use std::ffi::OsStr;
+use std::io::BufRead;
+use std::process::Command;
```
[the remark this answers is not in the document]
miraclx: `contract::build()` being an intuitive path is what informed the name of the file, but yeah, we can rename it.

cargo-near/src/abi/mod.rs
```diff
@@ -28,31 +30,43 @@ pub(crate) fn execute(
...
+            ("CARGO_PROFILE_DEV_OPT_LEVEL", "0"),
+            ("CARGO_PROFILE_DEV_DEBUG", "0"),
+            ("CARGO_PROFILE_DEV_LTO", "off"),
+        ],
+        util::dylib_extension(),
+    )?;
 
-    let cargo_toml = generation::generate_toml(manifest_path)?;
-    fs::write(near_abi_gen_dir.join("Cargo.toml"), cargo_toml)?;
+    // todo! experiment with reusing cargo-near for extracting data from the dylib
```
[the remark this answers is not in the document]
miraclx: Just a thought, nothing definitive. But I think we can skip the intermediate code generation for extracting data from the dylib to instead dynamically loading it at runtime by `cargo-near` and extracting data from it. Yet to experiment with it, but it is possible. Should make things a bit neater.

cargo-near/src/contract.rs
miraclx: Nice catch. And yeah, you're right.

austinabell:
High-level issues with these changes:
- Does not close #21 because it does not include `RUSTFLAGS='-C link-arg=-s'` with the build, doesn't have the output to the `near` sub-directory of target, doesn't implement the bundle output (part of this issue, we will do later)
- If res folder doesn't exist, it silently ignores (related to comment above I believe)
  - Behaviour should be that it creates the directory if it doesn't exist, and copy the output over
- Embed flow is embedding the JSON as pretty-formatted. There is no reason to include the whitespace in the Wasm binary, it's super inefficient (especially since right now the changes in the SDK are putting this in wasm memory which gets compiled)

cargo-near/src/contract.rs
```diff
@@ -0,0 +1,93 @@
...
+    if !invoke_rustup(&["target", "list", "--installed"])?
+        .lines()
+        .any(|target| target.as_ref().map_or(false, |t| t == COMPILATION_TARGET))
+    {
+        anyhow::bail!("rust target `{}` is not installed", COMPILATION_TARGET);
+    }
+
+    let crate_metadata = CrateMetadata::collect(CargoManifestPath::try_from(
+        args.manifest_path.unwrap_or_else(|| "Cargo.toml".into()),
+    )?)?;
```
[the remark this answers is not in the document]
miraclx: In this case, I think I prefer having this check outside of clap. Primarily because when reporting, clap prefixes the report with “Invalid value” which in most cases, would be a misnomer. As it won't be an invalid value, it just doesn't exist:

```console
error: Invalid value "a/Cargo.toml" for '--manifest-path <PATH>': manifest path does not exist
```

whereas, leaving that as-is, we can have errors like this:

```console
error: manifest path `a/Cargo.toml` does not exist
```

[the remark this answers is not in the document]
itegulov: I agree, this does look kind of weird. I wouldn't mind having this checked outside of clap.

cargo-near/src/lib.rs
```diff
@@ -25,31 +30,48 @@ pub enum NearCommand {
...
     pub doc: bool,
+    /// Do not generate ABI for the contract
+    #[clap(long, conflicts_with_all = &["doc", "embed-abi"])]
+    pub no_abi: bool,
+    /// Copy final artifacts to the this directory
+    #[clap(long, parse(from_os_str), value_name = "PATH")]
+    pub out_dir: Option<PathBuf>,
+    /// Path to the `Cargo.toml` of the contract to build
+    #[clap(long, parse(from_os_str), value_name = "PATH")]
+    pub manifest_path: Option<PathBuf>,
```
itegulov, line 69: I still feel like this should at least have default value set to `Cargo.toml` instead of unwrapping it in the middle of the code flow. This also has the benefit of showing up in the help command.

```suggestion
    #[clap(
        long,
        parse(from_os_str),
        default_value = "Cargo.toml",
        value_name = "PATH"
    )]
    pub manifest_path: PathBuf,
```

cargo-near/src/build.rs
```diff
@@ -0,0 +1,82 @@
...
+                "failed to copy `{}` to `{}`",
+                out_path.display(),
+                new_out_path.display(),
+            )
+        })?;
+        out_path = new_out_path;
+    }
+
+    // todo! if we embedded, check that the binary exports the __contract_abi symbol
+    println!("Contract successfully built at {}", out_path.display());
```
itegulov, line 79: Should we also print `abi.json` location if it was generated?

miraclx: Yeah, sounds logical. I'm also considering adding an equivalent to cargo's `--message-formatjson` in case people want to programmatically extract information from the build.

cargo-near/src/abi/mod.rs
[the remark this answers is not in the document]
miraclx: #35

cargo-near/src/lib.rs
miraclx: > This also has the benefit of showing up in the help command.

The description already hints at `Cargo.toml`.

```console
        --manifest-path <PATH>    Path to the `Cargo.toml` of the contract to build
```

This way, it's more consistent with `cargo build`:

```console
        --manifest-path <PATH>      Path to Cargo.toml
```

`cargo build`'s behavior without specifying a manifest path is to scan parent directories up to the root for a `Cargo.toml` and print this when it can't find one:

```console
error: could not find `Cargo.toml` in `/some/path/somewhere` or any parent directory
```

`cargo near`'s behavior right now is to check the current directory for a `Cargo.toml` and print this when it can't find one:

```console
error: manifest path `Cargo.toml` does not exist
```

cargo-near/src/util.rs
```diff
@@ -184,3 +185,28 @@ pub(crate) fn compile_project(
...
+    dir.canonicalize()
+        .with_context(|| format!("failed to access output directory `{}`", dir.display()))
+}
+
+/// Copy a file to a destination.
+///
+/// Does nothing if the destination is the same as the source to avoid truncating the file.
+pub(crate) fn copy(from: &Path, to: &Path) -> anyhow::Result<PathBuf> {
+    let out_path = to.join(from.file_name().unwrap());
+    if from.parent().unwrap() != to {
```
itegulov, line 202: Any reason not to just simply compare `from` and `out_path` here? Not a big deal either way, just one less unwrap.

cargo-near/src/build.rs
```diff
@@ -18,31 +17,27 @@ pub(crate) fn run(args: BuildCommand) -> anyhow::Result<()> {
...
-        .then(|| abi::write_to_file(&crate_metadata, args.doc))
-        .transpose()?;
+    let mut abi_path = None;
+    if !args.no_abi {
+        let AbiResult { path } = abi::write_to_file(&crate_metadata, args.doc)?;
+        abi_path.replace(util::copy(&path, &out_dir)?);
+    }
 
-    let wasm_artifact = if args.embed_abi {
+    let mut wasm_artifact = if let Some(ref abi_path) = abi_path {
```
itegulov, line 38: Wait, did you intend to get rid of the `args.embed_abi` check?

miraclx: Fixed.

itegulov (approved):

PR #33, opened 2022-08-25
feat: add binary serialization formats and compression when embedding

itegulov:
Still WIP, opening for visibility.

Table below shows some tests that I have done. Cell values are the size of the abi file after serialization+compression in bytes.
<table>
<tr>
	<td> ser+comp / contract
	<td> abi
	<td> ft
	<td> nft
<tr>
	<td> JSON (min)
	<td> 1694
	<td> 4834
	<td> 7148
<tr>
	<td> JSON (min)+snap
	<td> 672
	<td> 1265
	<td> 1590
<tr>
	<td> JSON (min)+zstd
	<td> 498
	<td> 914
	<td> 1113
<tr>
	<td> CBOR
	<td> 1379
	<td> 3991
	<td> 5842
<tr>
	<td> CBOR+snap
	<td> 602
	<td> 1239
	<td> 1580
<tr>
	<td> CBOR+zstd
	<td> 480
	<td> 909
	<td> 1132
</table>

### DIFF
--- a/cargo-near/src/abi.rs
+++ b/cargo-near/src/abi.rs
@@ -1,5 +1,6 @@
 use crate::cargo::{manifest::CargoManifestPath, metadata::CrateMetadata};
-use crate::{util, AbiCommand, AbiCompression, AbiFormat};
+use crate::{util, AbiCommand};
+use near_abi::AbiRoot;
 use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
@@ -10,12 +11,22 @@ pub(crate) struct AbiResult {
     pub path: PathBuf,
 }
 
-pub(crate) fn write_to_file(
+#[derive(Clone, Debug)]
+pub(crate) enum AbiFormat {
+    Json,
+    JsonMin,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum AbiCompression {
+    NoOp,
+    Zstd,
+}
+
+pub(crate) fn generate_abi(
     crate_metadata: &CrateMetadata,
     generate_docs: bool,
-    format: AbiFormat,
-    compression: AbiCompression,
-) -> anyhow::Result<AbiResult> {
+) -> anyhow::Result<AbiRoot> {
     fs::create_dir_all(&crate_metadata.target_directory)?;
 
     let original_cargo_toml: toml::value::Table =
@@ -55,6 +66,16 @@ pub(crate) fn write_to_file(
     if !generate_docs {
         strip_docs(&mut contract_abi);
     }
+
+    Ok(contract_abi)
+}
+
+pub(crate) fn write_to_file(
+    contract_abi: &AbiRoot,
+    crate_metadata: &CrateMetadata,
+    format: AbiFormat,
+    compression: AbiCompression,
+) -> anyhow::Result<AbiResult> {
     let near_abi_serialized = match format {
         AbiFormat::Json => serde_json::to_vec_pretty(&contract_abi)?,
         AbiFormat::JsonMin => serde_json::to_vec(&contract_abi)?,
@@ -119,8 +140,14 @@ pub(crate) fn run(args: AbiCommand) -> anyhow::Result<()> {
             .unwrap_or_else(|| crate_metadata.target_directory.clone()),
     )?;
 
+    let format = if args.no_pretty {
+        AbiFormat::JsonMin
+    } else {
+        AbiFormat::Json
+    };
+    let contract_abi = generate_abi(&crate_metadata, args.doc)?;
     let AbiResult { path } =
-        write_to_file(&crate_metadata, args.doc, args.format, args.compression)?;
+        write_to_file(&contract_abi, &crate_metadata, format, AbiCompression::NoOp)?;
 
     let abi_path = util::copy(&path, &out_dir)?;
 
--- a/cargo-near/src/abi.rs
+++ b/cargo-near/src/abi.rs
@@ -82,7 +82,10 @@ pub(crate) fn write_to_file(
     };
     let near_abi_compressed = match compression {
         AbiCompression::NoOp => near_abi_serialized,
-        AbiCompression::Zstd => zstd::encode_all(near_abi_serialized.as_slice(), 0)?,
+        AbiCompression::Zstd => zstd::encode_all(
+            near_abi_serialized.as_slice(),
+            *zstd::compression_level_range().end(),
+        )?,
     };
 
     let out_path_abi = crate_metadata.target_directory.join(format!(
--- a/cargo-near/src/abi.rs
+++ b/cargo-near/src/abi.rs
@@ -79,7 +79,7 @@ fn abi_file_extension(format: AbiFormat, compression: AbiCompression) -> &'stati
         AbiCompression::NoOp => match format {
             AbiFormat::Json | AbiFormat::JsonMin => "json",
         },
-        AbiCompression::Zstd => "bin",
+        AbiCompression::Zstd => "zst",
     }
 }
 
--- a/cargo-near/src/abi.rs
+++ b/cargo-near/src/abi.rs
@@ -1,5 +1,5 @@
 use crate::cargo::{manifest::CargoManifestPath, metadata::CrateMetadata};
-use crate::{util, AbiCommand};
+use crate::{util, AbiCommand, AbiCompression, AbiFormat};
 use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
@@ -13,6 +13,8 @@ pub(crate) struct AbiResult {
 pub(crate) fn write_to_file(
     crate_metadata: &CrateMetadata,
     generate_docs: bool,
+    format: AbiFormat,
+    compression: AbiCompression,
 ) -> anyhow::Result<AbiResult> {
     fs::create_dir_all(&crate_metadata.target_directory)?;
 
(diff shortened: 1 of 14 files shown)


austinabell:
Interesting, if easy to compare, I'd wonder what `JSON+zstd` or some compression would look like in comparison to `CBOR+zstd`. If the compressed bytes is only slightly larger, it would minimize dependencies needed to decode it (avoid pulling in cbor if not necessary).

itegulov:
> I'd wonder what JSON+zstd or some compression would look like in comparison to CBOR+zstd

Added those tests to the table and yeah it seems like zstd covers most of the benefit CBOR provides us. Probably not worth to bother with CBOR for a few bytes of benefit (~1% better on average).

itegulov:
So, I couple of things to say about this PR:

1. I have experimented with borsh, but unfortunately it is really hard to derive borsh ser/de for JsonSchema types. Borsh does not have any remote derivation capabilities and also does not expose any of the derivation logic. Manually implementing ser/de traits for all JSON Schema types is a big chore (there are a lot of types...). So realistically we would have to copy JSON Schema models from either schemafy or schemars (which might not be a bad idea in general tbh) and derive ser/de in our crate.
2. Calling `cargo near build` will result into:
```
Contract Successfully Built!
   - Binary: /home/itegulov/Work/near/near-sdk-rs/examples/abi/res/abi.wasm
   -    ABI: /home/itegulov/Work/near/near-sdk-rs/examples/abi/res/abi_abi.bin
```

Which is most likely not what the user wanted. I see three ways we could go here. First option: we do not expose `--format` and `--compression` and just assume that everyone wants pretty json for `cargo near abi`, pretty json for `cargo near build` output and (optionally) minified json+zstd for `--embed-abi`. Second option: we introduce two different set of options `--format`/`--compression` and `--embed-format`/`--embed-compression`. Third option: a mix of both, we leave `--format`/`--compression`, but always assume minified json+zstd for embedding. WDYT?

cargo-near/src/abi.rs
```diff
@@ -53,15 +55,34 @@ pub(crate) fn write_to_file(
...
+        AbiFormat::Json => serde_json::to_vec_pretty(&contract_abi)?,
+        AbiFormat::JsonMin => serde_json::to_vec(&contract_abi)?,
+    };
+    let near_abi_compressed = match compression {
+        AbiCompression::NoOp => near_abi_serialized,
+        AbiCompression::Zstd => zstd::encode_all(near_abi_serialized.as_slice(), 0)?,
+    };
+
+    let out_path_abi = crate_metadata.target_directory.join(format!(
+        "{}_abi.{}",
```
miraclx, line 68: When building a contract, and embedding abi, this exports a compressed ABI file. Shouldn't ABI compression should be limited to contract embedding only, or wdyt? So the exported ABI file should be the same as if you ran `cargo near abi`, which is a pretty-printed json file.

cargo-near/src/abi.rs
```diff
@@ -53,15 +55,34 @@ pub(crate) fn write_to_file(
...
 
     Ok(AbiResult { path: out_path_abi })
 }
 
+fn abi_file_extension(format: AbiFormat, compression: AbiCompression) -> &'static str {
+    match compression {
+        AbiCompression::NoOp => match format {
+            AbiFormat::Json | AbiFormat::JsonMin => "json",
+        },
+        AbiCompression::Zstd => "bin",
```
miraclx, line 82: Any reason for not using `.zst`?

cargo-near/src/lib.rs
```diff
@@ -70,6 +76,24 @@ pub struct BuildCommand {
     /// Path to the `Cargo.toml` of the contract to build
     #[clap(long, parse(from_os_str), value_name = "PATH")]
     pub manifest_path: Option<PathBuf>,
+    /// Serialization format to use
+    #[clap(long, value_enum, default_value = "json-min")]
+    pub format: AbiFormat,
+    /// Compression algorithm to use
+    #[clap(long, value_enum, default_value = "zstd")]
+    pub compression: AbiCompression,
```
miraclx, line 84: Do we plan to support more than `zstd`? Shouldn't we standardize it so tools that extract ABI can know which algorithm they'd need to decompress the view call response? So maybe this should just be `--uncompressed`?

cargo-near/src/abi.rs
miraclx: Just saw this was mentioned in your comment.

itegulov: See my comment above. It is not clear to me what the best flow would be here, seems opinionated.

itegulov: Honestly, makes me inclined not to expose `--format`/`--compression` for now and postpone this decision until we understand the use cases a little better.

cargo-near/src/abi.rs
itegulov: Nothing really, just wasn't aware that there is a special extension.

cargo-near/src/lib.rs
itegulov: > Do we plan to support more than zstd?

Probably not, but I want to keep our options open just in case. It would be easier to migrate from zstd to snap (hypothetically) if we can have support for both in intermediary versions until we deprecate and delete support for zstd.

cargo-near/src/lib.rs
```diff
@@ -47,6 +47,12 @@ pub struct AbiCommand {
     /// Path to the `Cargo.toml` of the contract to build
     #[clap(long, parse(from_os_str), value_name = "PATH")]
     pub manifest_path: Option<PathBuf>,
+    /// Serialization format to use
+    #[clap(long, value_enum, default_value = "json")]
+    pub format: AbiFormat,
```
miraclx, line 52: Maybe this should be `--pretty`?

cargo-near/src/abi.rs
miraclx:
```console
$ unzst res/abi_abi.bin
zstd: res/abi_abi.bin: unknown suffix (.zst/.tzst/.gz/.tgz/.lzma/.xz/.txz/.lz4/.tlz4 expected).
Can't derive the output file name. Specify it with -o dstFileName. Ignoring.
```

cargo-near/src/lib.rs
austinabell: Agree, if someone really needed to tweak manually themselves, they could just fork this. Better that this is kept as streamlined as possible to make it more easily consumed by tooling

cargo-near/src/abi.rs
itegulov: Already changed it to `.zst`, but that is a really good point. Haven't tried `unzst` manually.

cargo-near/src/abi.rs
miraclx: Opinionated, but I think.

1. We assume JSON as the ABI format. So there's no need for `--format`, seems redundant. Instead, we add `--pretty`.
2. We assume, when embedding, we always want to optimize for size. So compression is enabled by default. But we leave `--uncompress` to opt-out.

itegulov: > --pretty

Did you mean `--no-pretty` (or `--ugly` :grimacing:) or do you propose to minify json by default?

> So compression is enabled by default. But we leave --uncompress to opt-out.

Does it even makes sense to opt-out if all our tooling is going to expect this to be compressed?

austinabell (commented):
I just wonder if, even when embedding, a developer would still want to have the standard ABI file generated. Otherwise, they would have to do it manually, when it seems like very little lift to just generate the ABI in the regular format.

cargo-near/src/abi.rs
miraclx: Let's just leave both out. And only have a format toggle on the abi command.

cargo-near/src/abi.rs
```diff
@@ -53,15 +55,34 @@ pub(crate) fn write_to_file(
...
-        .target_directory
-        .join(crate_metadata.root_package.name.replace('-', "_") + "_abi.json");
-    fs::write(&out_path_abi, near_abi_json)?;
+    let near_abi_serialized = match format {
+        AbiFormat::Json => serde_json::to_vec_pretty(&contract_abi)?,
+        AbiFormat::JsonMin => serde_json::to_vec(&contract_abi)?,
+    };
+    let near_abi_compressed = match compression {
+        AbiCompression::NoOp => near_abi_serialized,
+        AbiCompression::Zstd => zstd::encode_all(near_abi_serialized.as_slice(), 0)?,
```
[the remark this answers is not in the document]
itegulov: Nope, thanks for pointing it out. I have tried out max level (22) compression and it shaves FT ABI size from 914 bytes down to 826 bytes. zstd documentation says that >=20 levels require more memory, but I wouldn't expect this to matter in our case, so we should be safe to use it. Just for a reference point, with level 15 FT ABI size is 830 bytes. So there is a minuscule difference, but it might of course differ on a case-to-case basis.

cargo-near/src/abi.rs
itegulov: Agreed, got rid of it

cargo-near/src/lib.rs
```diff
@@ -47,6 +47,9 @@ pub struct AbiCommand {
     /// Path to the `Cargo.toml` of the contract to build
     #[clap(long, parse(from_os_str), value_name = "PATH")]
     pub manifest_path: Option<PathBuf>,
+    /// Do not prettify ABI (will generate minified JSON)
+    #[clap(long)]
+    pub no_pretty: bool,
```
miraclx, line 52: idk, this doesn't seem very intuitive to me. what do you think of `--minify` or `--compact`?

austinabell: Unclear, seems opinionated. with `--compact` that is ambiguous whether it's referring to the ABI or the contract. Another option to throw in the ring is `--compact-abi`? `--no-pretty` still seems fine.

Or we could just remove altogether and if someone needs to compact it, they just use some external tool to compact it themselves, or we just wait until someone asks for this feature.

itegulov: Googled `--no-pretty` and there are a couple of well-known projects that use it, but overall it is indeed pretty rare. I am just going with `--compact-abi`. Seems the least opinionated, but we can always change while in early semver it if there is a strong reason for it.

austinabell (approved):

miraclx (approved):

PR #35, opened 2022-08-26
abi: dynamically load the dylib at runtime

miraclx:
Based off #23

Instead of building an intermediate project that loads the `dylib` to extract data, we can dynamically load the dylib at runtime from inside `cargo-near` itself.

https://github.com/near/cargo-near/pull/31#discussion_r953578045


itegulov (commented):
Awesome, I was unaware that it was this easy to load and call a dylib in runtime!

This change has one unintended side-effect though: we no longer create `target/near` directory automatically, so writing ABI to `target/near/abi.json` can fail if the directory is missing. Try running `cargo clean && cargo near build --out-dir res --release --embed-abi` on the `abi` example from SDK repo to see what I mean.

miraclx:
That is correct, right, nice catch!

itegulov (approved):

PR #39, opened 2022-08-27
use shared target folder for faster CI

miraclx:
Tracking issue: https://github.com/near/cargo-near/issues/41

Supersedes #24.

Current integration tests generate isolated projects with distinct target folders despite compiling ~99% of the same artifacts.

Causing current CI runs to run upwards of an hour.

With this patch, CI completes in under 10 minutes.

---

Based on #35, because in my tests, leaving the codegen in caused inconsistent results, ~didn't bother investigating why. https://github.com/near/cargo-near/runs/8050886022?check_suite_focus=true~

I eventually looked into it. See #40 for an alternative that merges straight into the main branch.

### DIFF
--- a/cargo-near/src/abi.rs
+++ b/cargo-near/src/abi.rs
@@ -4,8 +4,6 @@ use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
 
-const ABI_FILE: &str = "abi.json";
-
 /// ABI generation result.
 pub(crate) struct AbiResult {
     /// Path to the resulting ABI file.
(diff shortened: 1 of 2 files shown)


cargo-near/src/abi.rs
```diff
@@ -54,7 +52,9 @@ pub(crate) fn write_to_file(
         strip_docs(&mut contract_abi);
     }
     let near_abi_json = serde_json::to_string(&contract_abi)?;
-    let out_path_abi = crate_metadata.target_directory.join(ABI_FILE);
+    let out_path_abi = crate_metadata
+        .target_directory
+        .join(crate_metadata.root_package.name.replace('-', "_") + "_abi.json");
```
miraclx, line 57: This PR also prepends the package name to the name of the ABI file.

If you'd prefer to leave the ABI file as `abi.json`, there's another path I can take to arrive at the same result without changing the filename.

lmk what you think.

itegulov (approved):

cargo-near/src/abi.rs
itegulov: I think this is fine, no need for #42

austinabell (approved):

PR #45, opened 2022-08-30
feat: cover `cargo near abi` options with tests

itegulov:


integration-tests/tests/cargo/mod.rs
```diff
@@ -44,10 +44,10 @@ fn test_dependency_local_path_with_version() -> anyhow::Result<()> {
     let near_sdk_dep_path = near_sdk_dir.path().join("near-sdk");
 
     // near-sdk = { path = "::path::", version = "4.1.0-pre.2", features = ["abi"] }
-    let abi_root = generate_abi_fn! {
-        with Cargo "/templates/sdk-dependency/_Cargo_local_path_with_version.toml",
-        and vars HashMap::from([("path", near_sdk_dep_path.to_str().unwrap())]);
-
+    let abi_root = generate_abi_fn_with! {
+        Cargo: "/templates/sdk-dependency/_Cargo_local_path_with_version.toml";
+        Env: HashMap::from([("path", near_sdk_dep_path.to_str().unwrap())]);
```
miraclx, line 49: `Env` here is a bit misleading, don't you think? You'd have to either know beforehand, or read the macro code to know what it actually is.

miraclx (approved):
lgtm.

integration-tests/tests/cargo/mod.rs
itegulov: True. Changed it to `Vars:`

PR #46, opened 2022-08-30
update near-sdk in integration tests to pre-released version

miraclx:
This should effectively match the state it was in before ABI PRs were merged in. But with the version update.

### DIFF
--- a/integration-tests/templates/_Cargo.toml
+++ b/integration-tests/templates/_Cargo.toml
@@ -7,13 +7,9 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
+near-sdk = { version = "4.1.0-pre.3", features = ["abi"] }
 serde = { version = "1", features = ["derive"] }
 schemars = "0.8"
-
-[dependencies.near-sdk]
-git = "https://github.com/near/near-sdk-rs.git"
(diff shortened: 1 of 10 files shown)


austinabell (approved):
Is there the goal to support `near-sdk = "4.1.0-pre.3"` without the ABI feature? Assumed that was the point of cargo-near enabling the `__abi-generate` when generating the ABI which enables `abi`?

miraclx:
We're nearly there. I suppose the only blocker for that is finalizing https://github.com/near/near-sdk-rs/pull/891.

For now, the use of the `abi` feature gives access to the `NearSchema` derive. And since `cargo-near` can't do anything without those derives, it requires that feature flag.

But once `NearSchema` gets merged in, we can remove the `abi` feature.

austinabell:
> We're nearly there. I suppose the only blocker for that is finalizing [near/near-sdk-rs#891](https://github.com/near/near-sdk-rs/pull/891).
> 
> For now, the use of the `abi` feature gives access to the `NearSchema` derive. And since `cargo-near` can't do anything without those derives, it requires that feature flag.
> 
> But once `NearSchema` gets merged in, we can remove the `abi` feature.

Not necessarily, because you don't need the derive if all interface types already implement the schema types. Also, you could just not use the `NearSchema` convenience derive, that's optional, is it not?

PR #47, opened 2022-08-31
feat: cover `cargo near build` with tests

itegulov:
Caught one bug with `--doc` requiring `--embed-abi`, but otherwise seems to work as intended.


integration-tests/src/lib.rs
```diff
@@ -92,3 +96,97 @@ macro_rules! generate_abi_fn {
...
+        $(let opts = format!("cargo near build {}", $cli_opts);)?;
+        let result_dir = $crate::invoke_cargo_near! {
+            $(Cargo: $cargo_path;)? $(Vars: $cargo_vars;)?
+            Opts: opts;
+            Code:
+            $($code)*
+        };
+
+        let manifest_dir: std::path::PathBuf = env!("CARGO_MANIFEST_DIR").into();
+        let workspace_dir = manifest_dir.parent().unwrap().join("target").join("_abi-integration-tests");
```
miraclx, line 120: \+ `std::env::set_var("CARGO_TARGET_DIR", workspace_dir.join("target"));`

itegulov: Why do I need this here? This snippet is needed purely to decide where the resulting wasm file is located.

integration-tests/src/lib.rs
```diff
@@ -92,3 +96,97 @@ macro_rules! generate_abi_fn {
...
+        } else {
+            None
+        };
+
+        $crate::BuildResult { wasm, abi_root, abi_compressed }
+    }};
+}
+
+#[macro_export]
+macro_rules! build {
```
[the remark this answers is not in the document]
itegulov: Yeah, left it here to stay consisent with `generate_abi!` and it might be useful in the future. Can delete if you want to, though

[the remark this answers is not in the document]
miraclx: No strong opinion, just wasn't sure if I missed something.

integration-tests/src/lib.rs
miraclx: Ah, right, `build_with!` invokes `invoke_cargo_near!`. Ignore this.

miraclx (approved):

Issue #29, opened 2022-08-18
Bad error message for incorrect dependency


austinabell:
The error message, if someone has the `near-sdk` dependency as anything like `near-sdk = "4.0.0"` or `near-sdk = "4.1.0-pre.1`, the error message is:

```
ERROR: near-sdk dependency should be a table
```

Which is very unhelpful and does not indicate why this is failing. Ideally, for previous versions it should indicate what the minimum SDK version supported is, and for ones that don't apply the `abi` feature, it should either indicate that it is required or be able to add that feature automatically.

cc @shelegdmitriy

austinabell:
Adding this as another error message that is bad:

the package `non-fungible-token` depends on `near-sdk`, with features: `__abi-generate` but `near-sdk` does not have these features.

This depends on https://github.com/near/near-sdk-rs/pull/893 (which hasn't currently been merged or released) which will be unusable unless you know that.

cc @miraclx

PR #48, opened 2022-08-31
feat: improve error messages

itegulov:
Fixes https://github.com/near/cargo-near/issues/29


cargo-near/src/abi.rs
```diff
@@ -32,19 +33,32 @@ pub(crate) fn generate_abi(
...
-        .ok_or_else(|| anyhow::anyhow!("near-sdk dependency not found"))?
-        .as_table()
-        .ok_or_else(|| anyhow::anyhow!("near-sdk dependency should be a table"))?
-        .get("features")
-        .and_then(|features| features.as_array())
-        .map(|features| features.contains(&toml::Value::String("abi".to_string())))
-        .unwrap_or(false)
-    {
-        anyhow::bail!("Unable to generate ABI: NEAR SDK \"abi\" feature is not enabled")
+        .and_then(|deps| deps.get("near-sdk"))
```
miraclx, line 38: This won't catch
 - `near-sdk = { version = "..", features = ["abi"] }`
 - `near = { package = "near-sdk", version = ".." }`

and outright fail for `near-sdk = { git = ".." }`

miraclx (changes requested):
I'm taking a different approach of checking if the `near-sdk` dependency expects the `__abi-generate` and `__abi-embed` features. Instead of discriminating by any source specifier.

cargo-near/src/abi.rs
itegulov: > near-sdk = { version = "..", features = ["abi"] }

What do you mean? This is absolutely being caught. We have multiple existing tests covering this and similar scenarios.

> near = { package = "near-sdk", version = ".." }

This, most likely, is never (or at least not in the foreseeable future) going to be supported. The problem lies in `near-sdk` itself which expects the crate to be called `near_sdk`. See https://github.com/near/cargo-near/blob/main/integration-tests/tests/cargo/mod.rs#L137

> and outright fail for near-sdk = { git = ".." }

No, it won't. You can only specify a git dependency in a *table* which is handled in a different branch.

itegulov:
> I'm taking a different approach of checking if the near-sdk dependency expects the __abi-generate and __abi-embed features. Instead of discriminating by any source specifier.

I mean there is nothing stopping us from combining the two approaches. Directly checking for versions might lead to better contextual errors (depends on how you implement your approach, obviously I haven't seen it yet), and then you could check for `__abi-generate`.

cargo-near/src/abi.rs
miraclx: > What do you mean? This is absolutely being caught. We have multiple existing tests covering this and similar scenarios.

It's actually not. From the current logic, when the SDK dep is a table, it skips version checking altogether. And only checks for the `abi` feature.

> No, it won't. You can only specify a git dependency in a table which is handled in a different branch.

Same situation.

Since the `abi` feature flag is no longer required to generate ABI, when this check gets removed, many false negatives would scale through this pretty easily.

And when the dep is a table with version, one could specify "4.0" and it won't report any error until it hits the first `cargo metadata` which reports a verbose error:

```toml
[dependencies]
near-sdk = { version = "4", features = ["abi"] }
```

```console
error: Error invoking `cargo metadata`. Your `Cargo.toml` file is likely malformed

Caused by:
    `cargo metadata` exited with an error: error: failed to select a version for `near-sdk`.
        ... required by package `adder v0.1.0 (/Users/miraclx/git/near/near-sdk-rs/examples/adder)`
    versions that meet the requirements `^4` (locked to 4.0.0) are: 4.0.0
    
    the package `adder` depends on `near-sdk`, with features: `abi` but `near-sdk` does not have these features.
    
    
    failed to select a version for `near-sdk` which could resolve this conflict
```

> This, most likely, is never (or at least not in the foreseeable future) going to be supported. The problem lies in near-sdk itself which expects the crate to be called near_sdk. See [https://github.com/near/cargo-near/blob/main/integration-tests/tests/cargo/mod.rs#L137](https://github.com/near/cargo-near/blob/main/integration-tests/tests/cargo/mod.rs?rgh-link-date=2022-08-31T08%3A57%3A14Z#L137)

fair point.

itegulov: Ok, waiting for your PR then

miraclx: #49

PR #49, opened 2022-08-31
add check for supported near-sdk version

### LINKED ISSUE #29: Bad error message for incorrect dependency

miraclx:
Fixes #29

Reduced the scope of errors from dep checking to 3.

1. <code> unable to appropriately resolve the dependency graph, perhaps your \`Cargo.toml\` file is malformed </code>
2. <code> \`near-sdk\` dependency not found </code>
3. <code> unsupported \`near-sdk\` version. expected 4.1.* or higher </code>

Hopefully they're descriptive enough.

Adapted the integration tests from #48.


cargo-near/src/abi.rs
```diff
@@ -27,24 +27,35 @@ pub(crate) fn generate_abi(
...
+                .packages
+                .iter()
+                .find(|pkg| pkg.id == near_sdk.pkg)
+        })
+        .ok_or_else(|| anyhow::anyhow!("`near-sdk` dependency not found"))?;
+
+    for required_feature in ["__abi-generate", "__abi-embed"] {
+        if !near_sdk_dep.features.contains_key(required_feature) {
+            anyhow::bail!("unsupported `near-sdk` version. expected 4.1.* or higher");
+        }
```
itegulov, line 58: I must have missed where this conversation hapenned, but do we no longer require `abi` feature to be enabled? My only concern is this might be confusing to users that can successfully generate ABI using `cargo-near`, but who are not aware that there is an `abi` feature they can enable to get access to `NearSchema` and potentially some other stuff.

miraclx: https://github.com/near/cargo-near/pull/46

`cargo-near` can already build ABI without that requirement if all the interface types already implement `JsonSchema` or `BorshSchema`.

Ideally, we should get to a point where even `NearSchema` isn't feature-gated as well.

itegulov: Yes, I saw that, but I am asking more generally how we came to this decision of removing the `abi` feature.

> Ideally, we should get to a point where even NearSchema isn't feature-gated as well.

The problem with this is that it is unlikely to stabilize soon. `abi` is acting kind of like `unstable` in our case.

itegulov (commented):
Also, you can remove `#[ignore]` from `cargo::test_dependency_platform_specific`. Seems like `cargo metadata` is choosing the right dependency depending on the target which is a cool side effect.

itegulov (approved):
In any case, the changes look good to me even as is. Going to approve to unblock.

cargo-near/src/abi.rs
miraclx: > decision of removing the abi feature.

Just the `abi` feature _requirement_, where unnecessary.

- If you're already using features gated behind the `abi` feature, in spite of `cargo-near`, you already have the feature enabled otherwise `cargo build` doesn't pass anyway.
- And, if you're not, but all the types you do use already have `NearSchema` derived, there's no need to force that requirement.

austinabell: @itegulov do you have a reason why an ABI should not be allowed to be generated from a crate unless it uses the `abi` feature on the SDK dependency?

I see it as a benefit since a lot of devs coming in do not know how to enable features, so having the requirement, even if all of their types are already compatible, to generate the ABI seems like a little bit of a hurdle.

miraclx: Pulled this out into #50 so it's handled independently of this patch.

itegulov: No, I think it is fine to get rid of `abi`. I was just unsure of what to do with `NearSchema`. Moving under `unstable` makes sense to me.

itegulov (approved):

PR #50, opened 2022-08-31
remove abi feature requirement

miraclx:
Context: #49.


itegulov:
I and @austinabell discussed this during NEARCon and we kind of agreed that the best scenario here might be just enabling `abi` feature by default in SDK. The use pattern will be the same as in this PR, but also anyone who, for whatever reason, does not want to compile ABI codegen can just disable `abi` feature. What do you think?

miraclx:
Okay, this makes sense. I'll just test that the `abi` feature activation check holds for default features, and we should be good on this front.

austinabell:
> I and @austinabell discussed this during NEARCon and we kind of agreed that the best scenario here might be just enabling `abi` feature by default in SDK. The use pattern will be the same as in this PR, but also anyone who, for whatever reason, does not want to compile ABI codegen can just disable `abi` feature. What do you think?

For posterity, the motivation was around avoiding the bad experience of not having things fail during cargo check/RLS and only during `cargo build`, which would make it harder to debug (or require enabling the feature manually on contract). Open to a contrary opinion, though. It's not ideal that we add this functionality to the existing framework since it is a breaking change for contracts that have custom types in the interface.

miraclx:
Closed in favor of #85!

PR #51, opened 2022-08-31
chore: release 0.2.0

miraclx:
First pre-release of `cargo-near`

Also removed the author line from the help output, matching `neard`'s format.


austinabell (approved):

.gitignore
```diff
@@ -1,7 +1,9 @@
 # Build files
 /target
-Cargo.lock
```
itegulov, line 3: You are probably already aware of this, but `cargo install` ignores lockfile when executed without `--locked`. Nothing actionable here, just thought I would note.

README.md
```diff
@@ -1,18 +1,53 @@
+<!-- markdownlint-disable MD041 -->
```
itegulov, line 1: Why is this necessary? Don't we have a top-level heading in the first line?

itegulov (approved):

.gitignore
miraclx: Yeah, that inclusion is for git.

README.md
miraclx: typo. `MD041` -> `MD014`

Issue #38, opened 2022-08-27
CI: network errors on macOS CI runs


miraclx:
Opening this for visibility.

Been seeing this issue come up again and again in the macOS CI runs. And it doesn't happen in the Linux runs.

![CleanShot 2022-08-27 at 18 28 59@2x](https://user-images.githubusercontent.com/16881812/187035840-0147279e-f30c-428f-be32-3e837cb82c45.png)

At the moment, we're compiling 39 generated examples as part of our integration tests. And in many cases, at least one would have this error, causing the whole job to fail.

Of course, you can retry and, often, that works, but it's an issue nonetheless.

It seems the error might originate from `git2`, cargo uses this internally, and [we do too](https://github.com/near/cargo-near/blob/885437e5c966766e981e36ccd9a88fcdeb6cf882/integration-tests/tests/cargo/mod.rs#L12).

<details>
<summary> In <a href="https://github.com/near/cargo-near/runs/8051012360?check_suite_focus=true#step:3:499">this CI run</a>, cargo included a suggestion that I'll investigate </summary>

```console
test cargo::test_dependency_explicit ... FAILED
Caused by:
    `cargo metadata` exited with an error:     Updating git repository `[https://github.com/near/near-sdk-rs.git`](https://github.com/near/near-sdk-rs.git%60)
    warning: spurious network error (2 tries remaining): SecureTransport error: connection closed via error; class=Net (12)
    warning: spurious network error (1 tries remaining): SecureTransport error: connection closed via error; class=Net (12)
    error: failed to get `near-sdk` as a dependency of package `test_dependency_explicit v0.1.0 (/Users/runner/work/cargo-near/cargo-near/target/_abi-integration-tests/test_dependency_explicit)`
    
    Caused by:
      failed to load source for dependency `near-sdk`
    
    Caused by:
      Unable to update https://github.com/near/near-sdk-rs.git?rev=03bab8db6145e038626c4b15b34b0bd8a33fba44
    
    Caused by:
      failed to clone into: /Users/runner/.cargo/git/db/near-sdk-rs-7ba52202f378a9d9
    
    Caused by:
      network failure seems to have happened
      if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
      https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
    
    Caused by:
      SecureTransport error: connection closed via error; class=Net (12)
```

</details>

If that works, we might have to favor the git CLI over the use of `git2`.

[...]

PR #57, opened 2022-10-11
chore: update readme

itegulov:
We should stick to the uniform style set by https://github.com/near/near-sdk-rs and https://github.com/near/workspaces-rs

Preview link: https://github.com/near/cargo-near/blob/daniyar/update-readme/README.md


miraclx (approved):
You finally experienced the issue in https://github.com/near/cargo-near/issues/38

PR #59, opened 2022-10-13
feat: switch to stderr and print build stages

### LINKED ISSUE #37: Add output describing different build stages, switch to stderr
Noticed with https://github.com/near/cargo-near/pull/36 we can improve the build process by adding logs indicating which part of the build process is happening, so it's a little less opaque.

Indicating the ABI generation step, build step, and anything else would be useful (also for debugging)

Also, noticed we have print statements, and to match the Rust output (which is to stderr), these added statements and the ones that already exist should use `eprintln!` rather than `println!`.

Not a big deal, but if someone is trying to redirect the output of the build step, it might be very awkward to work with stdout and stderr

itegulov:
Closes #37


cargo-near/src/util/step_printer.rs
```diff
@@ -0,0 +1,41 @@
...
+        if self.cur_step > self.max_steps {
+            panic!(
+                "Expected process to have exactly {} steps, but encountered an extra step #{}",
+                &self.max_steps, &self.cur_step
+            )
+        }
+        eprintln!(
+            " {} {}",
+            format!("[{}/{}]", &self.cur_step, &self.max_steps).bold(),
+            msg.green().bold()
```
miraclx, line 27: This is a little too strict, don't you think? ideally, we should be able to step print any colors at the caller site.

cargo-near/src/build.rs
```diff
@@ -68,7 +78,7 @@ pub(crate) fn run(args: BuildCommand) -> anyhow::Result<()> {
     wasm_artifact.path = util::copy(&wasm_artifact.path, &out_dir)?;
 
     // todo! if we embedded, check that the binary exports the __contract_abi symbol
-    println!("{}", "Contract Successfully Built!".green().bold());
+    step_printer.print_step(&format!("{}", "Contract Successfully Built!"));
```
miraclx, line 81: ... so this would either revert to:

```suggestion
    step_printer.print_step(&"Contract Successfully Built!".green().bold());
```

Otherwise it can be simplified into:

```suggestion
    step_printer.print_step("Contract Successfully Built!");
```

cargo-near/src/build.rs
```diff
@@ -43,19 +47,25 @@ pub(crate) fn run(args: BuildCommand) -> anyhow::Result<()> {
...
+            step_printer.print_step("Compressing ABI for embedding");
             let AbiResult { path } = abi::write_to_file(
                 &contract_abi,
                 &crate_metadata,
                 AbiFormat::JsonMin,
                 AbiCompression::Zstd,
             )?;
             min_abi_path.replace(util::copy(&path, &out_dir)?);
+        } else {
+            step_printer.print_step("Skip embedding");
```
miraclx, line 59:
```suggestion
            step_printer.print_step("Skipped embedding");
```

cargo-near/src/build.rs
```diff
@@ -43,19 +47,25 @@ pub(crate) fn run(args: BuildCommand) -> anyhow::Result<()> {
         pretty_abi_path.replace(util::copy(&path, &out_dir)?);
 
         if args.embed_abi {
+            step_printer.print_step("Compressing ABI for embedding");
```
miraclx, line 50:
```suggestion
            step_printer.print_step("Compressing ABI to be embedded");
```

cargo-near/src/build.rs
```diff
@@ -1,19 +1,23 @@
...
 use crate::cargo::{manifest::CargoManifestPath, metadata::CrateMetadata};
+use crate::util::StepPrinter;
 use crate::{abi, util, BuildCommand};
 use colored::Colorize;
 use std::io::BufRead;
 
 const COMPILATION_TARGET: &str = "wasm32-unknown-unknown";
 
 pub(crate) fn run(args: BuildCommand) -> anyhow::Result<()> {
+    let mut step_printer = StepPrinter::new(7);
```
miraclx, line 11: if `--no-abi` is activated, won't this affect the count?

cargo-near/src/util/step_printer.rs
itegulov: Well, the idea is to print uniform build step messages. Kind of like `cargo` prints `Updating`/`Compiling`/`Finished` etc in the same boldness and color. Do you think there is an argument for this to be customizable on the caller site?

cargo-near/src/build.rs
itegulov: You are right. Hmm, I thought having statically known build step count could be cool for progress visibility, but now I am thinking it might become a nightmare to support long term. WDYT? Worth bothering with?

cargo-near/src/util/step_printer.rs
miraclx: Was just thinking of something like `Skipped embedding` which seems more like a filler than an actual step. Making it green seems like it accomplishes something, whereas it doesn't.

cargo-near/src/build.rs
miraclx: Yeah, it'll be quite problematic. What do you think about this pattern:

```console
 [•] Checking Environment... [done]
 [•] Collecting cargo project metadata... [done]
 [•] Extracting ABI from contract
    ... cargo stderr ...
 [•] Compressing ABI to be embedded... [done]
 [•] Building Contract (with embedded ABI)
    ... cargo stderr ...
 [✓] Contract Successfully Built!
```

would be neat if we could also indent cargo `std{out,err}` output to follow the style.

itegulov: Played with the colors a bit and this is what I have so far. LMK what you think:

latest-screenshot

Indenting the cargo output turned out to be a little troublesome, but I think I have figured it out in the end. A bit annoying that they are aligning the step names by the right side, but oh well.

cargo-near/src/build.rs
```diff
@@ -78,7 +76,7 @@ pub(crate) fn run(args: BuildCommand) -> anyhow::Result<()> {
     wasm_artifact.path = util::copy(&wasm_artifact.path, &out_dir)?;
 
     // todo! if we embedded, check that the binary exports the __contract_abi symbol
-    step_printer.print_step("Contract Successfully Built!");
+    util::print_success("Contract successfully built!");
```
miraclx, line 79: +1 for standardizing the case, caps as-needed.

cargo-near/src/build.rs
```diff
@@ -47,27 +48,24 @@ pub(crate) fn run(args: BuildCommand) -> anyhow::Result<()> {
...
 
     if let (true, Some(abi_path)) = (args.embed_abi, &min_abi_path) {
         cargo_args.extend(&["--features", "near-sdk/__abi-embed"]);
         build_env.push(("CARGO_NEAR_ABI_PATH", abi_path.to_str().unwrap()));
-        step_printer.print_step("Building contract (with embedded ABI)");
-    } else {
-        step_printer.print_step("Building contract");
     }
-
+    util::print_step("Generating ABI");
```
miraclx, line 68:
```suggestion
    util::print_step("Building Contract");
```

cargo-near/src/util/mod.rs
```diff
@@ -73,25 +74,52 @@ where
 
     cmd.arg(command);
     cmd.args(args);
+    // Ensure that cargo uses color output for piped stderr.
+    cmd.args(["--color", "always"]);
```
miraclx, line 78: Ideally, we should expose this through our CLI too. Doesn't have to happen in this PR though.

`--color auto,always,never`

miraclx (commented):
Nice work!

cargo-near/src/build.rs
miraclx: Just realized in the comment above I commended you on standardizing the case, then went ahead to violate that in my this comment 😆

itegulov: Ahahaha, I didn't even notice myself

cargo-near/src/util/mod.rs
itegulov: https://github.com/near/cargo-near/issues/60

cargo-near/src/build.rs
miraclx: > A bit annoying that they are aligning the step names by the right side, but oh well.

Is there a chance we can align it to match cargo's? how would that even look.

miraclx (approved):

austinabell (approved):

itegulov:
@miraclx I am merging this, and we can pull your PRs directly to `main`

cargo-near/src/util/mod.rs
miraclx: Fixed in #86.

PR #63, opened 2022-10-14
alt build stages style

miraclx:
@itegulov, what do you think of this styling pattern?

CleanShot 2022-10-15 at 11 44 51@2x

### DIFF
--- a/cargo-near/src/util/mod.rs
+++ b/cargo-near/src/util/mod.rs
@@ -106,7 +106,7 @@ where
         let stderr_reader = BufReader::new(child_stderr);
         let stderr_lines = stderr_reader.lines();
         for line in stderr_lines {
-            eprintln!("  {}", line.expect("failed to read cargo stderr"));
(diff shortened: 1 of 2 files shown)


itegulov:
Yoo, looks clean. I guess we don't need to pad the cargo output as much then. And also would be nice to incorporate the last lines with output files in the style somehow too :thinking:

miraclx:
Yeah, didn't want to have to push the text farther right. Another thing I noticed is when there's a cargo error, it doesn't have any colors. And for some reason, it's written to stdout instead of stderr. So it doesn't have the indented scoping that the build process has.

itegulov:
> It's because of --message-format json, got it.

Yeah, I could not find a way to combine JSON message format with colored output.

miraclx:
> Yeah, I could not find a way to combine JSON message format with colored output.

Seems to be the current situation, anyway, the error example in #65 isn't so bad either.

PR #65, opened 2022-10-14
report cargo stdout errors immediately

miraclx:
Context: https://github.com/near/cargo-near/pull/63#issuecomment-1278800638

Skips the message parsing redundancy and stdout buffer.

CleanShot 2022-10-15 at 11 50 03@2x

### DIFF
--- a/cargo-near/src/util/mod.rs
+++ b/cargo-near/src/util/mod.rs
@@ -1,11 +1,11 @@
 use crate::cargo::manifest::CargoManifestPath;
 use anyhow::{Context, Result};
 use cargo_metadata::diagnostic::DiagnosticLevel;
-use cargo_metadata::Message;
+use cargo_metadata::{Artifact, Message};
 use std::collections::HashSet;
 use std::ffi::OsStr;
 use std::fs;
-use std::io::{BufRead, BufReader, Read};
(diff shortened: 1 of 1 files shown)


miraclx:
Before this, since `stderr` got reported before `stdout`, we had this neat race conflict where cargo talks about "previous errors" yet the errors come after.

CleanShot 2022-10-14 at 14 56 16@2x

itegulov (approved):
:+1:

cargo-near/src/util/mod.rs
```diff
@@ -96,11 +79,29 @@ where
...
     let thread_stdout = thread::spawn(move || {
-        let mut result = Vec::new();
-        child_stdout
-            .read_to_end(&mut result)
-            .expect("failed to read cargo stdout");
-        result
+        let mut artifacts = vec![];
+        let stdout_reader = std::io::BufReader::new(child_stdout);
+        Message::parse_stream(stdout_reader)
+            .map(|m| m.unwrap())
```
miraclx, line 85: hm, any need to panic here?

miraclx:
Turns out warnings aren't being logged to stdout currently, this fixes that. But raises #67.

itegulov (approved):

PR #68, opened 2022-10-14
Deduplicate warnings

### LINKED ISSUE #67: Duplicate warnings
Using `cargo near` to occasionally build a contract while in active development can be a bit noisy, since both the step that aids ABI extraction and the step that does the actual contract compilation both invoke rustc, which reports the same warnings.

It's different for errors because it fails at the first pass.

Only relevant after #65.

CleanShot 2022-10-15 at 11 57 47@2x

miraclx:
Fixes #67

Hides warnings at the ABI generation stage only when building contracts

| Before | After |
| :-: | :-: |
| CleanShot 2022-10-15 at 11 57 47@2x | CleanShot 2022-10-15 at 11 55 35@2x |

### DIFF
--- a/cargo-near/src/util/mod.rs
+++ b/cargo-near/src/util/mod.rs
@@ -158,7 +158,12 @@ pub(crate) fn compile_project(
     args: &[&str],
     env: Vec<(&str, &str)>,
     artifact_extension: &str,
+    hide_warnings: bool,
 ) -> anyhow::Result<CompilationArtifact> {
+    if hide_warnings {
+        env::set_var("RUSTFLAGS", "-Awarnings");
+    }
+
     let artifacts = invoke_cargo(
         "build",
         [&["--message-format=json"], args].concat(),
--- a/cargo-near/src/util/mod.rs
+++ b/cargo-near/src/util/mod.rs
@@ -156,12 +156,12 @@ pub struct CompilationArtifact {
 pub(crate) fn compile_project(
     manifest_path: &CargoManifestPath,
     args: &[&str],
-    env: Vec<(&str, &str)>,
+    mut env: Vec<(&str, &str)>,
     artifact_extension: &str,
     hide_warnings: bool,
 ) -> anyhow::Result<CompilationArtifact> {
     if hide_warnings {
-        env::set_var("RUSTFLAGS", "-Awarnings");
+        env.push(("RUSTFLAGS", "-Awarnings"));
     }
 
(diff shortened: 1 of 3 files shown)


cargo-near/src/util/mod.rs
```diff
@@ -161,7 +161,12 @@ pub(crate) fn compile_project(
     args: &[&str],
     env: Vec<(&str, &str)>,
     artifact_extension: &str,
+    hide_warnings: bool,
 ) -> anyhow::Result<CompilationArtifact> {
+    if hide_warnings {
+        env::set_var("RUSTFLAGS", "-Awarnings");
```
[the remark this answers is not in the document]
miraclx: 😆 trusted GitHub Copilot a little more than I should've 😆

miraclx:
@itegulov, when you get the time please can you test this. My one test in the screenshot worked, but ever since, I've been unable to get any warnings at all through `cargo near`, still investigating why. Thought it was a caching thing, but I've cleaned by target folder repeatedly and still nothing.

itegulov:
@miraclx yeah I am not getting any warnings either

miraclx:
Fixed in [`5378499` (#65)](https://github.com/near/cargo-near/pull/65/commits/53784994559e2ea82df559b1391317bba61169e4), merged in here.

itegulov (approved):
Can confirm that warnings appear now!

cargo-near/src/util/mod.rs
```diff
@@ -156,14 +157,38 @@ pub struct CompilationArtifact {
...
+                let rustflags: &mut String = final_env
+                    .entry(key)
+                    .or_insert_with(|| std::env::var(key).unwrap_or_default());
+                if !rustflags.is_empty() {
+                    rustflags.push(' ');
+                }
+                rustflags.push_str(value);
+            }
+            _ => {
+                final_env.insert(key, value.to_string());
```
austinabell, line 182: curious, why use a map instead of just a vec of `(K, V)` if you're not checking the duplication of keys on insert?

austinabell (approved):

cargo-near/src/util/mod.rs
miraclx: Well, actually, for `RUSTFLAGS`, we're checking to ensure we don't replace previous entries, but instead concatenate them. So it's here to allow us quickly index the map, instead of having to iterate the whole `Vec<(K, V)>`.

PR #69, opened 2022-10-15
report appropriate status

miraclx:
### Before

CleanShot 2022-10-15 at 17 42 28@2x

### After

CleanShot 2022-10-18 at 12 07 05@2x


itegulov (approved):
Makes sense to me. We can also replace `error: ...` with `✗ ...`, leaving this up to you.

PR #70, opened 2022-10-17
feat: enforce `near-abi` version match

### LINKED ISSUE #61: Add a check for `near-abi` version
We are moving fast and occasionally make breaking changes to the schema format. This means that using `cargo-near` and `near-sdk` with different underlying `near-abi` versions can lead to unpleasant user experience. Currently `cargo-near` just segfaults because it is trying to dynamically load and call the contract artifact with an unexpected `near-abi` version. We should add an explicit check instead.

CC @miraclx

itegulov:
Fixes #61

Just a dumb `MAJOR.MINOR` check for now, we can tweak it later if needed

### DIFF
--- a/cargo-near/src/abi.rs
+++ b/cargo-near/src/abi.rs
@@ -59,6 +59,19 @@ pub(crate) fn generate_abi(
         }
     }
 
+    if !crate_metadata
+        .root_package
+        .dependencies
+        .iter()
+        .find(|dep| dep.name == "near-sdk")
+        .ok_or_else(|| anyhow::anyhow!("`near-sdk` dependency not found"))?
+        .features
+        .iter()
+        .any(|feature| feature == "abi")
+    {
+        anyhow::bail!("`near-sdk` dependency must have the `abi` feature enabled")
+    }
+
     let near_sdk_node = crate_resolve
         .nodes
         .iter()
@@ -91,19 +104,6 @@ pub(crate) fn generate_abi(
         } else {
             anyhow::bail!("your project is using `near-abi` {}, which is older than the currently installed `cargo-near` allows; please update your `near-sdk` dependency", actual_abi_ver);
         }
-    }
-
-    if !crate_metadata
-        .root_package
-        .dependencies
-        .iter()
-        .find(|dep| dep.name == "near-sdk")
-        .ok_or_else(|| anyhow::anyhow!("`near-sdk` dependency not found"))?
-        .features
-        .iter()
-        .any(|feature| feature == "abi")
-    {
-        anyhow::bail!("`near-sdk` dependency must have the `abi` feature enabled")
     }
 
     let dylib_artifact = util::compile_project(
--- a/cargo-near/src/abi.rs
+++ b/cargo-near/src/abi.rs
@@ -1,5 +1,6 @@
 use crate::cargo::{manifest::CargoManifestPath, metadata::CrateMetadata};
 use crate::{util, AbiCommand};
+use cargo_metadata::Version;
 use near_abi::AbiRoot;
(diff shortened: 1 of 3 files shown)


itegulov:
Will add a test for this once we release near-sdk 4.1.0-pre.4 with near-abi 0.2.0

cargo-near/src/abi.rs
```diff
@@ -71,6 +72,40 @@ pub(crate) fn generate_abi(
...
+                .packages
+                .iter()
+                .find(|pkg| pkg.id == near_abi.pkg)
+        })
+        .ok_or_else(|| anyhow::anyhow!("expected `near-sdk` to provide `near-abi`, please report this issue at https://github.com/near/cargo-near"))?;
+
+    let expected_abi_ver = &Version::parse(env!("CARGO_NEAR_ABI_VERSION"))?;
+    let actual_abi_ver = &near_abi_dep.version;
+    if expected_abi_ver.major != actual_abi_ver.major
+        || expected_abi_ver.minor != actual_abi_ver.minor
```
miraclx, line 100: Hm, should we be testing with the crate version? or the `schema_version`?

itegulov: The reason why this check is needed in the first place is because `ChunkedAbiEntry` returned from the dylib artifact has to conform to the `ChunkedAbiEntry` used by `cargo-near`, which depends on how Rust ABI views them. For example I highly doubt that Rust would allow you to interpret:

```rust
pub struct AbiMetadata {
    pub name: Option<String>,
    pub other: HashMap<String, String>,
}
```
and
```rust
pub struct AbiMetadata {
    pub name: Option<String>,
    pub version: Option<String>,
    pub other: HashMap<String, String>,
}
```
as the same object even if they both can belong to the same `schema_version` (extending metadata object does not bump the schema version obviously).

If we were to switch back to JSON-based ABI chunk generation, then we would be able only to check the `schema_version`.

miraclx: Yeah I understand that, suggesting we could consider returning `serde_json::Value` from the chunks, instead of the strict type. Then we can leave it up to the deserialization implementation to do the validation. Or what do you think?

But as-is works too.

itegulov: Hmm, that could work. In general, do you know what invariants you can depend on when working with low-level Rust ABI? For example, is it reasonable to assume that `serde_json::Value` is going to have the same ABI for all `major.minor` releases? Also, would I have to re-compile cargo-near whenever I install a new version of rustc?

miraclx: Not sure, honestly, best path to avoid inconsistencies later down the line would probably be to return JSON strings. And manually handle deserialization wherein we can validate the `schema_version`.

itegulov: Yeah, that's what I am thinking too right now. Any ideas on how to get `schema_version` from a crate though? Can we add like a custom cargo metadata key or something?

miraclx: So, `ChunkedAbiEntry` already includes `schema_version`.

Which means we can only do the compatibility check after the `Generating ABI` step.

miraclx: I especially like that `ChunkedAbiEntry` already does this compatibility check on deserialization.

https://github.com/near/near-abi-rs/blob/368245073aa5bf919ee30f9e6f2246f68eee1b40/near-abi/src/private.rs#L11-L12

So we get this for free.

itegulov: True, I will prepare an alternative PR then

itegulov:
Closed in favor of https://github.com/near/cargo-near/pull/74

PR #72, opened 2022-10-18
CI: run matrix paths independently

miraclx:
GitHub's matrix behavior is a little weird. By default, when one path causes an execution to fail, GitHub auto-cancels all other executions from the same matrix.

Example: https://github.com/near/cargo-near/actions/runs/3271855547/attempts/2

CleanShot 2022-10-18 at 13 39 59@2x

Ideally, we should test each one independently, and get all the possible errors/successes all at once, instead of incrementally.


itegulov (approved):
Makes sense to me!

PR #73, opened 2022-10-19
feat: return JSON-based ABI generation

itegulov:
Fixes https://github.com/near/cargo-near/issues/61

Basically reverts https://github.com/near/cargo-near/pull/35 with some refactoring and better error messages

Also, this whole `invoke_cargo` shebang is getting out of hand. I am surprised there is no crate that encapsulates this tbh

### DIFF
--- a/cargo-near/src/abi/generation.rs
+++ b/cargo-near/src/abi/generation.rs
@@ -1,0 +1,165 @@
+use crate::cargo::manifest::CargoManifestPath;
+use crate::cargo::metadata::CrateMetadata;
+use anyhow::Context;
+use quote::{format_ident, quote};
+use std::path::{Path, PathBuf};
+use std::{collections::HashSet, fs};
+use toml::value;
+
+pub(crate) fn generate_toml(
+    manifest_path: &CargoManifestPath,
+    crate_metadata: &CrateMetadata,
+) -> anyhow::Result<String> {
+    let original_cargo_toml = fs::read_to_string(&manifest_path.path)?;
+    let original_cargo_toml: toml::value::Table = toml::from_str(&original_cargo_toml)?;
+
+    let mut near_sdk = original_cargo_toml
+        .get("dependencies")
+        .context("Cargo.toml [dependencies] section not found")?
+        .get("near-sdk")
+        .context("`near-sdk` dependency not found")?
+        .as_table()
+        .context("`near-sdk` dependency should be a table")?
+        .clone();
+
+    let cargo_toml = include_str!("../../templates/_Cargo.toml");
+    let mut cargo_toml: toml::value::Table = toml::from_str(cargo_toml)?;
+    let package = cargo_toml
+        .get_mut("package")
+        .context("Cargo.toml template [package] section not found")?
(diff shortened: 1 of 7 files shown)


cargo-near/src/abi/mod.rs
```diff
@@ -86,18 +99,65 @@ pub(crate) fn generate_abi(
...
+                .parent()
+                .unwrap()
+                .to_string_lossy()
+                .to_string(),
+        )],
+        stdout_fn,
+        stderr_fn,
+    )?;
+
+    // This will fail to serialize if there is a `schema_version` mismatch
```
miraclx, line 135:
```suggestion
    // This will fail to deserialize if there is a `schema_version` mismatch
```

cargo-near/src/abi/generation.rs
```diff
@@ -0,0 +1,165 @@
...
+    });
+
+    let main_rs = quote! {
+        extern "Rust" {
+            #(#near_abi_function_defs)*
+        }
+
+        fn main() -> Result<(), std::io::Error> {
+            let abi_entries: Vec<near_sdk::__private::ChunkedAbiEntry> = vec![#(#near_abi_function_invocations),*];
+            let contents = serde_json::to_string_pretty(&abi_entries)?;
```
miraclx, line 153: We can actually move this bit to the SDK, simplifying things. See alt: https://github.com/near/cargo-near/pull/74

cargo-near/src/util/mod.rs
```diff
@@ -31,19 +31,36 @@ pub(crate) const fn dylib_extension() -> &'static str {
...
+    Args: IntoIterator<Item = Arg>,
+    Arg: AsRef<OsStr>,
+    TPath: AsRef<Path>,
+    Env: IntoIterator<Item = (EnvKey, EnvVal)>,
+    EnvKey: AsRef<OsStr>,
+    EnvVal: AsRef<OsStr>,
+    StdoutFn: FnOnce(ChildStdout) -> Result<StdoutResult, std::io::Error> + Send + 'static,
+    StdoutResult: Send + 'static,
+    StderrFn: FnOnce(ChildStderr) -> Result<StderrResult, std::io::Error> + Send + 'static,
+    StderrResult: Send + 'static,
```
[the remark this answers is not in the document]
miraclx: I've seen some crates use `cargo` as a direct dependency, apparently it has a public facing API separate from the CLI, just never used it directly.

https://docs.rs/cargo

cargo-near/src/abi/generation.rs
itegulov: Not really, it may look similar but there is an important difference that I have touched upon here: https://github.com/near/cargo-near/pull/74#pullrequestreview-1147284614

miraclx: Yeah, I was thinking a Rust `String` might be safe, but since we don't have that guarantee, it's just best to switch to the C ABI.

itegulov:
Closed in favor of https://github.com/near/cargo-near/pull/74

PR #74, opened 2022-10-19
replace strict typed casting with deserialization

miraclx:
CleanShot 2022-10-19 at 14 05 58@2x

Alternative to #73, depends on a small patch to SDK – https://github.com/near/near-sdk-rs/pull/939.

### DIFF
--- a/cargo-near/src/util/mod.rs
+++ b/cargo-near/src/util/mod.rs
@@ -278,7 +278,8 @@ pub(crate) fn extract_abi_entries(
         for symbol in near_abi_symbols {
             let entry: libloading::Symbol<extern "C" fn() -> *const std::ffi::c_char> =
                 lib.get(symbol.as_bytes())?;
-            match serde_json::from_slice(std::ffi::CStr::from_ptr(entry()).to_bytes()) {
+            match serde_json::from_slice(std::ffi::CString::from_raw(entry() as *mut _).to_bytes())
+            {
                 Ok(entry) => entries.push(entry),
                 Err(err) => {
                     // unfortunately, we're unable to extract the raw error without Display-ing it first
--- a/cargo-near/src/util/mod.rs
+++ b/cargo-near/src/util/mod.rs
@@ -280,8 +280,9 @@ pub(crate) fn extract_abi_entries(
     unsafe {
         let lib = libloading::Library::new(dylib_path)?;
         for symbol in near_abi_symbols {
-            let entry: libloading::Symbol<fn() -> String> = lib.get(symbol.as_bytes())?;
-            match serde_json::from_str(&entry()) {
+            let entry: libloading::Symbol<extern "C" fn() -> *const std::ffi::c_char> =
+                lib.get(symbol.as_bytes())?;
+            match serde_json::from_str(std::ffi::CStr::from_ptr(entry()).to_str()?) {
                 Ok(entry) => entries.push(entry),
                 Err(err) => {
                     // unfortunately, we're unable to extract the raw error without Display-ing it first
--- a/cargo-near/src/util/mod.rs
+++ b/cargo-near/src/util/mod.rs
@@ -280,9 +280,26 @@ pub(crate) fn extract_abi_entries(
     unsafe {
         let lib = libloading::Library::new(dylib_path)?;
         for symbol in near_abi_symbols {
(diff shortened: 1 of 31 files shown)


itegulov (changes requested):
So, this has fewer problems than our current approach, but it still depends on Rust ABI not changing between the point when you compiled `cargo-near` and when you are invoking it (e.g. you used rustc 1.61 to compile `cargo-near`, but now you are using rustc 1.64 to compile your smart contract).

From what I understand Rust ABI is not stable, so I would not be comfortable with depending on it not changing, unless someone has more insight on this topic. CC @austinabell

itegulov:
But thinking about this a bit more, couldn't we avoid Rust ABI entirely and load this with `libloading::Symbol<extern "C" fn() -> String>` instead? We would have to change the way SDK exposes the symbol obviously, but that should not be a big deal.

miraclx:
@itegulov, yeah I came to the same conclusion, but to be safe, we can replace `String` entirely with `*const c_char`

itegulov:
Hmm, weird. Have you tried it? I am getting an empty file instead of ABI.

miraclx:
Yeah, it works on my machine. lol. Let me try to reproduce the error.

miraclx:
Yeah, it works. Although, for posterity, the empty file issue was fixed in https://github.com/near/cargo-near/pull/75.

miraclx:
This patch requires a version of `near-sdk` that uses `near-abi v0.2.0`. So we'll have to merge #55 before this.

cargo-near/src/util/mod.rs
```diff
@@ -276,9 +276,27 @@ pub(crate) fn extract_abi_entries(
...
+                        if let Some((line, col)) = rest.rsplit_once(" column ") {
+                            if line.chars().all(|c| c.is_numeric())
+                                && col.chars().all(|c| c.is_numeric())
+                            {
+                                err_str.truncate(msg.len());
+                                err_str.shrink_to_fit();
+                                anyhow::bail!(err_str);
+                            }
+                        }
+                    }
```
itegulov, line 296: Good call on truncating the "at line ... column ..." stuff. It looks weird when there is a schema version mismatch, and all other errors should not happen  Even if they do, line+column do not help much when you can't easily see the intermediary JSON value.

itegulov (commented):
This looks good to me, but lets wait for the SDK change to come in first

cargo-near/src/util/mod.rs
miraclx: Exactly! Weird that there is no way to get the core error from `serde_json::Error`, but this works too.

cargo-near/src/util/mod.rs
```diff
@@ -276,9 +276,28 @@ pub(crate) fn extract_abi_entries(
     unsafe {
         let lib = libloading::Library::new(dylib_path)?;
         for symbol in near_abi_symbols {
-            let entry: libloading::Symbol<fn() -> near_abi::__private::ChunkedAbiEntry> =
+            let entry: libloading::Symbol<extern "C" fn() -> *const std::ffi::c_char> =
                 lib.get(symbol.as_bytes())?;
-            entries.push(entry().clone());
+            match serde_json::from_slice(std::ffi::CString::from_raw(entry() as *mut _).to_bytes())
```
austinabell, line 281: We might want https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html here instead of going through CString now

miraclx: Yeah, already patched in #80. I'll merge in here. Also, it would be `Vec::from_raw_parts` otherwise we'd be leaking memory.

miraclx:
@itegulov, this is ready. What do you think? Can you test it out?

itegulov:
@miraclx hmm weird it is not working for me. Tried our classic `adder` example and got this:

latest-screenshot

miraclx:
Have you tried cleaning your target folder? Maybe cargo is returning an artifact from previous builds? In the meantime, I'll spin up a Linux machine and see if I can reproduce this.

itegulov:
@miraclx yep, sorry, must have had some leftover artifacts indeed. Works fine now!

itegulov (approved):
LGTM. Don't forget to update the git rev to the master commit though once the SDK PR is merged.

miraclx:
You got it, chief!

PR #75, opened 2022-10-19
fix same-file truncation issue

miraclx:
Apparently, the paths returned from `cargo_metadata` like `Metadata::{`[`target_directory`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Metadata.html#structfield.target_directory)`,`[`workspace_root`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Metadata.html#structfield.workspace_root)`}` aren't absolute.

This assumed guarantee causes the logic in [`util::copy`](https://github.com/near/cargo-near/blob/5633f998671df2521c9dc03f568ec0442da712ef/cargo-near/src/util/mod.rs#L240-L255) to be invalidated.

Treating the same file as if it were different, because one of the paths had `../..` in it.

This patch migrates to early canonization, to avoid this error coming up again.

Props to @itegulov, for discovering this. We should probably cover this in tests.

### DIFF
--- a/cargo-near/src/cargo/metadata.rs
+++ b/cargo-near/src/cargo/metadata.rs
@@ -15,15 +15,17 @@ pub(crate) struct CrateMetadata {
 impl CrateMetadata {
     /// Parses the contract manifest and returns relevant metadata.
     pub fn collect(manifest_path: CargoManifestPath) -> Result<Self> {
-        let (metadata, root_package) = get_cargo_metadata(&manifest_path)?;
+        let (mut metadata, root_package) = get_cargo_metadata(&manifest_path)?;
+        metadata.target_directory = metadata.target_directory.canonicalize_utf8()?;
+        metadata.workspace_root = metadata.workspace_root.canonicalize_utf8()?;
+
         let mut target_directory = metadata.target_directory.as_path().join("near");
 
         // Normalize the package and lib name.
         let package_name = root_package.name.replace('-', "_");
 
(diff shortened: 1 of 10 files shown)


itegulov (approved):

cargo-near/src/cargo/metadata.rs
```diff
@@ -15,15 +15,17 @@ pub(crate) struct CrateMetadata {
 impl CrateMetadata {
     /// Parses the contract manifest and returns relevant metadata.
     pub fn collect(manifest_path: CargoManifestPath) -> Result<Self> {
-        let (metadata, root_package) = get_cargo_metadata(&manifest_path)?;
+        let (mut metadata, root_package) = get_cargo_metadata(&manifest_path)?;
+        metadata.target_directory = metadata.target_directory.canonicalize_utf8()?;
+        metadata.workspace_root = metadata.workspace_root.canonicalize_utf8()?;
```
miraclx, line 20: if we're forcing utf8 here, we might as well change `--out-dir`, `--manifest-path` to support valid `utf8` inputs.

Neither `cargo` nor `rustc` support non-utf8 names/paths.

CleanShot 2022-10-19 at 17 10 42@2x


CleanShot 2022-10-19 at 17 07 30@2x

miraclx: #76

miraclx:
Once merged, #76 fixes the CI issue here.

PR #76, opened 2022-10-19
only accept UTF-8 inputs

miraclx:
Neither `cargo` nor `rustc` support non-utf8 names/paths.

| `rustc` | `cargo` |
| :-: | :-: |
| CleanShot 2022-10-19 at 17 10 42@2x | CleanShot 2022-10-19 at 17 07 30@2x |

And while `cargo-near` currently accepts non-utf8 inputs, it panics at a `fs` operation.

CleanShot 2022-10-19 at 17 46 31@2x

This patch pre-emptively returns a helpful error.

CleanShot 2022-10-19 at 17 49 13@2x

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -493,6 +493,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "borsh",
+ "camino",
  "cargo-near",
  "clap",
  "function_name",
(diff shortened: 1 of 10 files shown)


itegulov (approved):
Makes sense. I wasn't aware that there are separate types for UTF8 paths

PR #79, opened 2022-10-20
feat: consolidate all sdk version related constants in one place

itegulov:
less pain for when we upgrade SDK version


miraclx (approved):
Awesome, love the simplicity!

austinabell (approved):

PR #86, opened 2022-11-04
feat: dynamic colors

miraclx:
Observed a weird behavior where `cargo-near` disables colors when `stdout` is piped, even though we're printing to `stderr`.

Looked into the logic of the dependency `colored`, and it's pretty [rigidly tied to `stdout`](https://github.com/mackwic/colored/blob/11ffd20e7b5d1b48e4de7bb78ced26131ae5e114/src/control.rs#L107-L108). But thankfully, you can [override it](https://github.com/mackwic/colored/blob/11ffd20e7b5d1b48e4de7bb78ced26131ae5e114/src/control.rs#L119-L121).

The logic implemented here is in two parts.

1. The `NO_COLOR` env. https://no-color.org/. `colored` also uses this internally.
3. The `--color <auto|always|never>` which takes precedence over the env.

This configuration also applies to the cargo subprocesses.


miraclx:
Alternatively, we can switch from `colored` to something like `console` that lets you enable/disable colors depending on what `tty` you're writing to.

https://docs.rs/console/latest/console/struct.Style.html#method.for_stderr

Although the syntax would look a bit like `eprintln!("{}", style("hello").for_stderr().cyan())`

cargo-near/src/lib.rs
```diff
@@ -68,6 +76,56 @@ pub struct BuildCommand {
...
+                    if atty::is(atty::Stream::Stderr) {
+                        Ok(ColorPreference::Always)
+                    } else {
+                        Ok(ColorPreference::Never)
+                    }
+                }
+            },
+            "always" => Ok(ColorPreference::Always),
+            "never" => Ok(ColorPreference::Never),
+            _ => Err(format!("invalid value: {}", s)),
```
austinabell, line 110: Would be nice to indicate this is an "invalid color preference" or something like that, value is very generic

austinabell (approved):
change seems reasonable!

cargo-near/src/lib.rs
miraclx: Yep, sounds good to me.

austinabell: ah, interesting, wouldn't expect it to swallow the error. Optional for sure, but maybe saves some debug time of someone in the future :D

Issue #4, opened 2022-06-29
Support for Windows


itegulov:
Using `cargo-near` on Windows results in these errors:
```
error: couldn't read \\?\C:\Users\diteg\Work\near-sdk-rs\examples\abi\target\near\abi-near-abi-gen\target\release\build\crunchy-35e8e00adbbf080a\out/lib.rs: The filename, directory name, or volume label syntax is incorrect. (os error 123)
  --> C:\Users\diteg\.cargo\registry\src\github.com-1ecc6299db9ec823\crunchy-0.2.2\src\lib.rs:36:1
   |
36 | include!(concat!(env!("OUT_DIR"), "/lib.rs"));
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `crunchy` due to previous error
warning: build failed, waiting for other jobs to finish...
error: couldn't read \\?\C:\Users\diteg\Work\near-sdk-rs\examples\abi\target\near\abi-near-abi-gen\target\release\build\rustversion-19b9a7e40293d578\out/version.expr: The filename, directory name, or volume label syntax is incorrect. (os error 123)
   --> C:\Users\diteg\.cargo\registry\src\github.com-1ecc6299db9ec823\rustversion-1.0.7\src\lib.rs:185:30
    |
185 | const RUSTVERSION: Version = include!(concat!(env!("OUT_DIR"), "/version.expr"));
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this error originates in the macro `include` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `rustversion` due to previous error
error: couldn't read \\?\C:\Users\diteg\Work\near-sdk-rs\examples\abi\target\near\abi-near-abi-gen\target\release\build\reed-solomon-erasure-72d98e9b864bbed5\out/table.rs: The filename, directory name, or volume label syntax is incorrect. (os error 123)
 --> C:\Users\diteg\.cargo\registry\src\github.com-1ecc6299db9ec823\reed-solomon-erasure-4.0.2\src\galois_8.rs:3:1
  |
3 | include!(concat!(env!("OUT_DIR"), "/table.rs"));
[...]

PR #90, opened 2022-11-09
Automated Releases

miraclx:
This introduces a CI workflow that automates GitHub releases as well as publishing to crates.io.

### How to update versions

- Submit a PR that bumps the version of the `Cargo.toml` file.
- Ensure the `Cargo.lock` file is also updated.
- Move all entries under the `Unreleased` section of the `CHANGELOG.md` file to a versioned section.
- Once the PR is merged, the release happens!

### How it works

Unlike most workflows, this doesn't trigger on a tag push. Instead, the first job in the workflow checks for a version bump to the `Cargo.toml` file when merged to master.

If a version bump was detected, we check if that version has already been published to crates.io. If it hasn't, we publish the crate. The second part of this, is checking if the tag already exists on GitHub. If it doesn't, we proceed with the release process.

The release process builds binaries for multiple platforms, and makes them available for download from the final release page as archives, including hashes of each archive in the release note, in case users need to verify them. This also tags the contributors that made the release happen, links to the version of the crate on crates.io, and appropriately tags pre-release versions on GitHub.

### Now for some cases

- If the version has already been published to crates.io and released to GitHub
   CleanShot 2022-11-09 at 13 38 15@2x
- Final release page
   CleanShot 2022-11-09 at 11 59 37@2x
- File Hashes
   CleanShot 2022-11-09 at 11 59 59@2x
- Pre-release tag
   CleanShot 2022-11-09 at 12 28 55@2x

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -290,10 +290,12 @@ jobs:
         echo "CHANGELOG_LINE=${CHANGELOG_LINE}" >> $GITHUB_ENV
 
         HASH_LINE="<details>\n<summary>Verify the SHA256 checksums of the release artifacts</summary>\n\n| File | Hash |\n| --- | --- |"
-        for file in artifacts/*/*; do
-          HASH_LINE="${HASH_LINE}\n| [\`$(basename "${file}")\`](https://github.com/near/cargo-near/releases/download/${{ env.GH_PROJECT_VERSION }}/$file) | \`$(sha256sum "${file}" | cut -d' ' -f1)\` |"
+        pushd artifacts
+        for file in */*; do
+          HASH_LINE="${HASH_LINE}\n| [\`${file}\`](https://github.com/near/cargo-near/releases/download/${{ env.GH_PROJECT_VERSION }}/$file) | \`$(sha256sum "${file}" | cut -d' ' -f1)\` |"
         done
         HASH_LINE="${HASH_LINE}\n\n</details>"
+        popd
         echo "HASH_LINE<<EOF" >> $GITHUB_ENV
         echo -e "${HASH_LINE}" >> $GITHUB_ENV
         echo "EOF" >> $GITHUB_ENV
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -254,14 +254,14 @@ jobs:
             https://api.github.com/repos/${{ github.repository }}/releases/generate-notes \
             -H 'Authorization: token ${{ github.token }}' \
             -d '{"tag_name":"${{ env.GH_PROJECT_VERSION }}","target_commitish":"${{ github.sha }}"}' \
-          | jq .body
+          | jq -r .body
         )"
         echo "Default Release Note"
         echo "${DEFAULT_RELEASE_NOTE}"
         echo
         CONTRIBUTORS="$(
-          echo "${DEFAULT_RELEASE_NOTE}" \
-          | perl -ne 'print if s/.+ by (@[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}) in https.+$/\1/i' \
+          <<< "${DEFAULT_RELEASE_NOTE}" \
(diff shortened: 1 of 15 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,322 @@
...
+    # ? - { name: 'linux-arm-gnu'     , target: arm-unknown-linux-gnueabihf , os: ubuntu-latest , use-cross: true }
+    # ? - { name: 'linux-arm-musl'    , target: arm-unknown-linux-musleabihf, os: ubuntu-latest , use-cross: true }
+    # ? - { name: 'linux-i686-gnu'    , target: i686-unknown-linux-gnu      , os: ubuntu-latest , use-cross: true }
+    # ? - { name: 'linux-i686-musl'   , target: i686-unknown-linux-musl     , os: ubuntu-latest , use-cross: true }
+    # ? - { name: 'win32-msvc'        , target: i686-pc-windows-msvc        , os: windows-latest,                 }
+        - { name: 'linux-x86_64-gnu'  , target: x86_64-unknown-linux-gnu    , os: ubuntu-latest ,                 }
+        - { name: 'linux-x86_64-musl' , target: x86_64-unknown-linux-musl   , os: ubuntu-latest , use-cross: true }
+        - { name: 'macos-x86_64'      , target: x86_64-apple-darwin         , os: macos-latest  ,                 }
+        - { name: 'win64-x86_64-gnu'  , target: x86_64-pc-windows-gnu       , os: windows-latest,                 }
+        - { name: 'win64-x86_64-msvc' , target: x86_64-pc-windows-msvc      , os: windows-latest,                 }
```
miraclx, line 163: This leads to https://github.com/near/cargo-near/issues/4, I haven't tested `cargo-near` builds on Windows lately.

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+      PROJECT_NAME: ${{ needs.check.outputs.project_name }}
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Install rust
+      uses: dtolnay/rust-toolchain@stable
+
+    - uses: Swatinem/rust-cache@v2
```
austinabell, line 122: cache seems unnecessary for the publish step

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+
+        if echo "${RELEASED_VERSIONS}" | grep -q "^v${PROJECT_VERSION}$"; then
+          echo "Project version [${PROJECT_VERSION}] has already been released."
+          echo "SHOULD_RELEASE_BINARY=false" >> $GITHUB_ENV
+        else
+          echo "Project version [${PROJECT_VERSION}] has not been released."
+          echo "SHOULD_RELEASE_BINARY=true" >> $GITHUB_ENV
+        fi
+
+  crate:
```
austinabell, line 108: seems like this should be named `publish`?

austinabell (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+
+    - uses: actions/upload-artifact@v3
+      with:
+        name: ${{ env.PKG_PATH }}
+        path: ${{ env.PKG_PATH }}
+        retention-days: 1
+
+  github:
+    runs-on: ubuntu-latest
+    needs: [check, binary]
```
miraclx, line 217: Worth highlighting that the release on GitHub doesn't depend on the crate release.

Yet it includes the crate link in the release notes.

What do you think? Should we wait for the crate to be published, or not?

.github/workflows/release.yml
miraclx: Indeed, it was helpful for testing, but unnecessary here!

itegulov:
This is very nice. Feels like it deserves to be its own independent Github Action. @miraclx have you considered this?

Also, the release process is valuable knowledge, so lets put it in some visible place like `RELEASE.md`.

PR #103, opened 2023-09-15
feat: Added interactive-clap for cargo-near

FroVolod:


cargo-near/src/lib.rs
```diff
@@ -1,136 +0,0 @@
-use std::{env, str::FromStr};
```
frol, line 1: lib.rs should stay as we want to make use of cargo-near commands in near-workspaces-rs, similarly to how we use near-cli-rs here. Please, move most of things except main() function back from main.rs to here.

cargo-near/src/main.rs
```diff
@@ -1,27 +1,106 @@
...
+    loop {
+        match <Cmd as interactive_clap::FromCli>::from_cli(
+            Some(cli.clone()),
+            global_context.clone(),
+        ) {
+            interactive_clap::ResultFromCli::Ok(cli_cmd)
+            | interactive_clap::ResultFromCli::Cancel(Some(cli_cmd)) => {
+                eprintln!(
+                    "Your console command:\n{} {}",
+                    std::env::args().next().as_deref().unwrap_or("./cargo-near"),
```
frol, line 84: We need to make sure that if users call `cargo near build` we don't suggest them to use `cargo-near near build`. I feel it is currently the case, correct me if I am wrong

cargo-near/src/abi_command/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Include rustdocs in the ABI file
+    #[interactive_clap(long)]
+    pub doc: bool,
+    /// Generate compact (minified) JSON
+    #[interactive_clap(long)]
+    pub compact_abi: bool,
+    /// Copy final artifacts to this directory
+    #[interactive_clap(long)]
+    #[interactive_clap(skip_default_input_arg)]
+    pub out_dir: Option<crate::types::utf8_path_buf::Utf8PathBuf>,
```
frol, line 16: I don't know the reason behind the use of Utf8PathBuf, so I would love to try and see if PathBuf would work just fine ([we already have it near-cli-rs](https://github.com/near/near-cli-rs/blob/master/src/types/path_buf.rs))

P.S. Obviously, try applying it everywhere we currently have this Utf8PathBuf and if everything works fine, remove the dependency from Cargo.toml as well.

frol (commented):
Nice! We are on the right path to getting `cargo near deploy`!

cargo-near/src/lib.rs
FroVolod: Done

cargo-near/src/main.rs
FroVolod: Done

frol (changes requested):
Please, refactor the project to have a similar structure to near-cli-rs / bos-cli-rs

frol (approved):
I believe it is ready to replace the previous implementation!

PR #113, opened 2023-10-13
feat: New command - deploy

### LINKED ISSUE #107: `cargo near deploy` command to deploy the current project to blockchain
`cargo near deploy` should do the following:

1. Run `cargo near build` code internally
2. Integrate with `near-cli-rs` code in such a way that interactive part picks up from `near contract deploy my-new-dev-account.testnet use-file ./target/near/adder.wasm ...`

So the final command could look like:

```
cargo near deploy \
    frol.near \
    with-init-call \
        new \
        json-args '{"param1": 1}' \
        prepaid-gas '100.000 TeraGas' \
        attached-deposit '0 NEAR' \
    network-config testnet \
    sign-with-...
```

```
cargo near deploy \
    frol.near \
    without-init-call \
    network-config testnet \
    sign-with-...
```

Note: We don't want to implement `dev-deploy` magic in cargo-near, instead, I suggest to introduce `cargo near create-dev-account` #106 

Note: All the features available in near-cli-rs for deployment should be inherited here, so near-cli-rs should be reused here.

FroVolod:
Resolves #107


cargo-near/src/commands/deploy/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+                    embed_abi: cli_build_command_args.embed_abi,
+                    doc: cli_build_command_args.doc,
+                    no_abi: cli_build_command_args.no_abi,
+                    out_dir: cli_build_command_args.out_dir.clone(),
+                    manifest_path: cli_build_command_args.manifest_path.clone(),
+                    color: cli_build_command_args.color.clone(),
+                }
+            } else {
+                build_command::BuildCommand::default()
+            };
```
frol, line 77: It seems that something like this should work:

```suggestion
        let build_command_args = clap_variant.build_command_args.map(BuildCommand::from_cli).unwrap_or_default();
```

cargo-near/src/commands/deploy/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    #[interactive_clap(skip_default_input_arg)]
+    #[interactive_clap(flatten)]
+    /// Specify a build command args:
+    build_command_args: build_command::BuildCommand,
+    #[interactive_clap(skip_default_input_arg)]
+    /// What is the contract account ID?
+    contract_account_id: near_cli_rs::types::account_id::AccountId,
+    #[allow(dead_code)]
+    #[interactive_clap(skip)]
+    file_path: crate::types::utf8_path_buf::Utf8PathBufInner,
```
frol, line 19: Looks like we need to use context here

frol (changes requested):

cargo-near/src/commands/deploy/mod.rs
FroVolod: Done

cargo-near/src/commands/deploy/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+impl Contract {
+    #[allow(dead_code)]
+    fn input_build_command_args(
+        _context: &near_cli_rs::GlobalContext,
+    ) -> color_eyre::eyre::Result<Option<build_command::BuildCommand>> {
+        Ok(None)
+    }
```
frol, line 124: Given that `from_cli` is implemented manually, and it does not call `input_build_command_args`, I believe you can remove this function.

FroVolod: Done

PR #125, opened 2024-01-24
refactor: Upgraded NEAR crates to 0.20.0 release

FroVolod:


cargo-near/Cargo.toml
```diff
@@ -43,11 +43,11 @@ linked-hash-map = { version = "0.5", features = ["serde_impl"] }
...
-near-primitives = "0.17.0"
-near-jsonrpc-client = "0.6.0"
-near-jsonrpc-primitives = "0.17.0"
+interactive-clap = "0.2.8"
+interactive-clap-derive = "0.2.8"
+near-cli-rs = { version = "0.7.7" }
+near-crypto = "0.20.0"
+near-primitives = "0.20.0"
+near-jsonrpc-client = "0.8.0"
+near-jsonrpc-primitives = "0.20.0"
```
frol, line 52: It seems that cargo-near does not use these. Can you check if we can remove these and any of the other dependencies currently listed here?

frol (approved):

PR #144, opened 2024-04-09
feat(source-scan): logging, refactor, image with same `cargo-near` embedded, remove `--no-docker` flag

dj8yfo:
Following output was generated by using the branch of current pr with [build local `cargo-near` Dockerfile](https://github.com/dj8yfo/cargo-near-image/tree/0.x.x-builer-local-cargo-near) => [image](https://hub.docker.com/r/dj8yfo/sourcescan/tags) on this [repo](https://github.com/dj8yfo/sample_no_workspace/tree/set_metadata)

Screenshot_20240412_230234
Screenshot_20240412_230436
Screenshot_20240412_230749

### DIFF
--- a/cargo-near/src/commands/build_command/docker.rs
+++ b/cargo-near/src/commands/build_command/docker.rs
@@ -113,9 +113,7 @@ impl super::BuildCommand {
             );
             println!(
                 "{}",
-                format!(
-                    "You can choose to opt out into non-docker build behaviour by using `--no-docker` flag.",
-                ).cyan()
+                "You can choose to opt out into non-docker build behaviour by using `--no-docker` flag.".cyan()
             );
 
             Err(color_eyre::eyre::eyre!(
--- a/cargo-near/src/commands/build_command/docker.rs
+++ b/cargo-near/src/commands/build_command/docker.rs
@@ -111,6 +111,12 @@ impl super::BuildCommand {
                     docker_cmd
                 ).yellow()
             );
+            println!(
+                "{}",
+                format!(
+                    "You can choose to opt out into non-docker build behaviour by `export NEAR_BUILD_ENVIRONMENT_REF=old`",
+                ).cyan()
+            );
 
             Err(color_eyre::eyre::eyre!(
                 "Reproducible build in docker container failed"
--- a/cargo-near/src/commands/build_command/docker.rs
+++ b/cargo-near/src/commands/build_command/docker.rs
@@ -114,7 +114,7 @@ impl super::BuildCommand {
             println!(
                 "{}",
                 format!(
-                    "You can choose to opt out into non-docker build behaviour by `export NEAR_BUILD_ENVIRONMENT_REF=old`",
+                    "You can choose to opt out into non-docker build behaviour by using `--no-docker` flag.",
                 ).cyan()
             );
 
--- a/cargo-near/src/commands/build_command/docker.rs
+++ b/cargo-near/src/commands/build_command/docker.rs
@@ -1,0 +1,222 @@
+use std::{
+    process::{id, Command, ExitStatus},
+    time::{SystemTime, UNIX_EPOCH},
+};
+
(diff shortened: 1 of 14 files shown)


cargo-near/src/commands/build_command/mod.rs
```diff
@@ -1,30 +1,18 @@
...
-
-use crate::{
-    types::{manifest::CargoManifestPath, metadata::CrateMetadata},
-    util,
-};
-
-pub mod build;
+mod build;
+mod docker;
+pub const INSIDE_DOCKER_ENV_KEY: &str = "NEAR_BUILD_ENVIRONMENT_REF";
```
frol, line 7: Given that the latest naming according to the semi-approved [NEP-330 extension is `build_info.build_environment`](https://github.com/near/NEPs/pull/533/files#diff-4955342da1cab11a6777c89e1ae874f185c844101af84775c413b87098aa9a12R104) and we use [`CARGO_*` variables in `near-sdk-rs`](https://github.com/near/near-sdk-rs/blob/master/near-sdk-macros/src/core_impl/contract_metadata/mod.rs#L51-L58), let's name it `CARGO_NEAR_BUILD_ENVIRONMENT`.

We will also need to pass `CARGO_NEAR_SOURCE_CODE_SNAPSHOT`, `CARGO_NEAR_CONTRACT_PATH`, and `CARGO_NEAR_BUILD_COMMAND`, so near-sdk-rs can pick those up and include in the metadata.

cargo-near/src/commands/build_command/docker.rs
```diff
@@ -0,0 +1,226 @@
...
+            util::handle_step("Parsing and validating `Cargo.toml` metadata...", || {
+                metadata::ReproducibleBuild::parse(&crate_metadata)
+            })?;
+
+        util::print_step("Running docker command step...");
+        let (status, docker_cmd) = self.docker_subprocess_step(docker_build_meta, &cloned_repo)?;
+
+        if status.success() {
+            util::print_success("Running docker command step (finished)");
+            // TODO: make this a `ClonedRepo` `copy_artifact` method
```
frol, line 69: TODO (please, keep track of all TODOs)

cargo-near/src/commands/build_command/docker.rs
```diff
@@ -0,0 +1,226 @@
...
+                        let mut contract_path = cloned_repo.contract_path.clone();
+                        contract_path.push("contract.wasm");
+                        contract_path
+                    };
+                    std::fs::copy::<std::path::PathBuf, camino::Utf8PathBuf>(
+                        entry.path(),
+                        wasm_path.clone(),
+                    )?;
+
+                    // TODO: ensure fresh
```
frol, line 92: TODO (please, keep track of all TODOs)

cargo-near/src/commands/build_command/docker.rs
```diff
@@ -0,0 +1,226 @@
...
+                .tmp_repo
+                .workdir()
+                .wrap_err("Could not get the working directory for the repository")?
+                .to_string_lossy()
+        );
+        let docker_container_name = format!("cargo-near-{}-{}", timestamp, pid);
+        let docker_image = docker_build_meta.concat_image();
+        let near_build_env_ref = format!("NEAR_BUILD_ENVIRONMENT_REF={}", docker_image);
+
+        // Platform-specific UID/GID retrieval
```
frol, line 168: This is a pretty pointless comment. More useful comment would be to explain that we only really need to match uid/gid on Linux since Docker is running natively and mounted FS needs to have proper permissions.

frol (approved):
I approve this PR as is. It targets a still-in-progress `source-scan-integration` branch, where we will review and refactor everything at once after we have a solid MVP

cargo-near/src/commands/build_command/docker.rs
dj8yfo: transferred to local task-tracking system

PR #157, opened 2024-04-23
feat(source-scan): build command in metadata

dj8yfo:
`build_command` key set: https://github.com/dj8yfo/sample_no_workspace/tree/metadata_image_repo_build_cmd
Screenshot_20240423_182924

---
`build_command` key not set: https://github.com/dj8yfo/sample_no_workspace/tree/metadata_image_repo
Screenshot_20240423_183205

---
typo in key: https://github.com/dj8yfo/sample_no_workspace/tree/metadata_image_repo_wrong_key
Screenshot_20240423_180816

### DIFF
--- a/cargo-near/src/commands/build_command/docker.rs
+++ b/cargo-near/src/commands/build_command/docker.rs
@@ -131,7 +131,8 @@ impl super::BuildCommand {
             "-c",
         ];
 
-        let cargo_cmd = self.compute_build_command(docker_build_meta.build_command.clone())?;
+        let cargo_cmd =
+            self.compute_build_command(docker_build_meta.container_build_command.clone())?;
         println!(" {} {}", "build command in container:".green(), cargo_cmd);
 
         docker_args.push(&cargo_cmd);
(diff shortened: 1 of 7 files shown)


cargo-near/src/main.rs
```diff
@@ -67,6 +67,7 @@ fn main() -> CliResult {
     } else {
         env::args().next().unwrap_or("./cargo".to_string())
     };
+    export_nep_330_build_command();
```
frol, line 70: I don't feel that it is the right place for it. Can we move it to the build command implementation?

cargo-near/src/commands/build_command/docker.rs
```diff
@@ -189,4 +162,82 @@ impl super::BuildCommand {
...
+        if self.no_release {
+            cargo_args.push("--no-release");
+        }
+
+        let color = self
+            .color
+            .clone()
+            .unwrap_or(crate::common::ColorPreference::Auto)
+            .to_string();
+        cargo_args.extend(&["--color", &color]);
```
frol, line 237: I don't see much value in passing `--color auto` explicitly when it was not specified. Can we conditionally add it there?

```suggestion
        if let Some(color) = self.color {
            cargo_args.extend(&["--color", &color.to_string()]);
        }
```

frol (approved):

PR #159, opened 2024-04-26
feat(source-scan): git checks, git metadata

dj8yfo:
Summary of changes:
- `git dirty` check copied to `cargo near build` flow, it produces a warning, unlike the error in `cargo near deploy` flow
- added `source_code_git_url` field to reproducible build metadata
- `git pushed to remote` check logic reworked
- repo is opened and `HEAD` is determined
- relative path of contract within repo is determined
- `CARGO_NEAR_SOURCE_CODE_SNAPSHOT` (adapted from [cargo::core::SourceId](https://docs.rs/cargo/latest/cargo/core/struct.SourceId.html)) and `CARGO_NEAR_CONTRACT_PATH` are exported into container
- [x] github, package in root of a repo: https://github.com/dj8yfo/sample_no_workspace/tree/6c5ae44150ce51349a95e1605ba053c4fe53cf28
- [x] github, package in subpath in repo : https://github.com/dj8yfo/sample_workspace/tree/b2d5580037c84fd4038b467564a9be9e520611de/self-updates/update
- [x] github, package in submodule in repo : https://github.com/dj8yfo/sample_workspace_with_submodules/tree/3d87c81517e1e23a5716c300fe4789b1aafc4c6f/self-updates
- [x] bitbucket, package in root of a repo: https://bitbucket.org/dj8yfomule/scratch_check_git/src/master/

Screenshot_20240503_204916


dj8yfo:
```bash
near contract call-function as-read-only sample-crate.testnet contract_source_metadata json-args {} network-config testnet now
```
=> 
```json
{
  "build_info": {
    "build_command": [
      "cargo",
      "near",
      "build",
      "--no-release",
      "--no-doc"
    ],
    "build_environment": "dj8yfo/sourcescan:0.x.x-dev-cargo-near-build-cmd-repo@sha256:6d4f090a8bb606bbddf396809d31b6c96a45d0bdd3e93d9e4bf538ff463246e9",
    "contract_path": null,
    "source_code_snapshot": "git+https://github.com/dj8yfo/sample_no_workspace.git?rev=05067bc12b8a8d811883b6b12fc87ba9bce9b103"
  },
  "link": "https://github.com/dj8yfo/sample_no_workspace/tree/05067bc12b8a8d811883b6b12fc87ba9bce9b103",
  "standards": [
    {
      "standard": "nep330",
      "version": "1.2.0"
    }
  ],
  "version": "0.7.7"
}

```

cargo-near/src/commands/build_command/docker.rs
```diff
@@ -83,64 +89,61 @@ impl super::BuildCommand {
...
-        let pid = id().to_string();
-        let timestamp = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap()
-            .as_secs()
-            .to_string();
+        let mut docker_cmd: Command = {
+            // Platform-specific UID/GID retrieval
+            #[cfg(unix)]
+            let uid_gid = format!("{}:{}", getuid(), getgid());
```
frol, line 95: It should be Linux-specific since we don't need to match the uid/gid on MacOS or other *nixes. Please, update the comment to highlight the reason for this mapping is that on Linux the volume is mounted natively and thus the unprivileged user inside Docker container should be able to write to the mounted folder that has the host user permissions

cargo-near/src/commands/build_command/docker.rs
```diff
@@ -239,4 +242,180 @@ impl super::BuildCommand {
...
+            .clone()
+            .and_then(|url| url.host_str().map(ToString::to_string))
+            .filter(|host| *host == "github.com")
+            .and(url);
+        let commit_hint = url.and_then(|url| {
+            let existing_path = url.path();
+            url.join(&format!("{}/tree/{}", existing_path, self.revision))
+                .ok()
+        });
+        commit_hint.map(|url| url.to_string())
```
frol, line 419: I believe this code is easier to read:

```suggestion
        let url = url::Url::parse(&self.repo_link?).ok()?;

        if url.host_str() == "github.com" {
            let existing_path = url.path();
            Some(url.join(&format!("{}/tree/{}", existing_path, self.revision))
                .ok()?
                .to_string())
        } else {
            None
        }
```

cargo-near/src/commands/build_command/docker.rs
dj8yfo: fixed by https://github.com/near/cargo-near/pull/159/commits/d8598dff6172b96c18b9db2135407e7d67362c89

cargo-near/src/commands/build_command/docker.rs
dj8yfo: fixed by https://github.com/near/cargo-near/commit/db93a3c1e8fe2cdefcd77f46efc886e17b7b0ad4

frol (approved):

PR #160, opened 2024-05-03
Support passing feature flags to `cargo` invocation

encody:
Adds support for passing feature flags to `cargo`.

## Motivation

Smart contract developers will sometimes use [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#the-features-section) to enable/disable certain portions of code during compilation (e.g. enable certain debugging features for the testnet version which are not available in the production version).

## Usage

```txt
cargo near build --features feat1,feat2
```

```txt
cargo near build --no-default-features
```

## Implementation Notes

It does no validation of the input received to the flag, transparently forwarding it as an opaque string directly to `cargo`, after appending `near-sdk/__abi-embed` if necessary.

### DIFF
--- a/cargo-near/src/commands/build_command/build.rs
+++ b/cargo-near/src/commands/build_command/build.rs
@@ -86,7 +86,7 @@ pub fn run(args: super::BuildCommand) -> color_eyre::eyre::Result<util::Compilat
         cargo_args.push(&cargo_feature_flags);
     }
 
-    if args.no_default_features.unwrap_or(false) {
+    if args.no_default_features {
         cargo_args.push("--no-default-features");
     }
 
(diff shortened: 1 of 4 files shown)


frol (approved):
@encody Nice! Looks good to me. @dj8yfo Do you have any objections for merging it (given the context of #134)

frol:
@encody Hey, @dj8yfo flagged that this PR forgot to pass the feature flags to the ABI generation code:

https://github.com/near/cargo-near/blob/358ce04ba0383e2a837c1c6fc99ec512379b061e/cargo-near/src/commands/abi_command/abi.rs#L87-L89

Could you contribute the fix, so this feature is complete?

PR #161, opened 2024-05-05
Fix: Also pass feature flags to ABI build step

encody:
Addresses https://github.com/near/cargo-near/pull/160#issuecomment-2094153730


frol (approved):

frol (approved):
@encody Thank you for addressing it! I resolved the merge conflict and approving it now

PR #168, opened 2024-07-02
Updates near-cli-rs and cargo-near to latest versions

elliotBraem:
`cargo near new` creates a new contract from template with pre-populated git workflows.

It seems the versions of near-cli-rs and cargo-near are no longer compatible, getting error:

```
Error: 
   0: expected `schema_version` to be ~0.3, but got 0.4.0: consider upgrading near-abi to a newer version
```

This pull request updates these two packages to latest versions. I've validated a successful deploy after modified workflows here: https://github.com/NEARBuilders/gateway/pull/457#issuecomment-2204429774


frol (approved):
@elliotBraem thank you for catching that!

PR #169, opened 2024-07-03
test: add docker build integration test

dj8yfo:
this [line](https://github.com/near/cargo-near/blob/main/integration-tests/src/lib.rs#L59) was fluke-nuking
the added test, 

due to following commands

```bash
export CARGO_TARGET_DIR=/home/target
cargo metadata
```

corrupting `cargo metadata` output,
so the new test was put into a separate binary.
Screenshot_20240703_203008


Canvinus (approved):

cargo-near/src/commands/build_command/docker.rs
```diff
@@ -129,7 +129,10 @@ impl super::BuildCommand {
                     &container_paths.crate_path,
                 ];
 
-                if std::env::var(SERVER_DISABLE_INTERACTIVE).is_err() {
+                log::info!("input device is a tty: {}", atty::is(atty::Stream::Stdin));
```
frol, line 132: Should it be removed or at least downgraded to debug?

integration-tests/tests/abi/borsh_schema.rs
```diff
@@ -7,6 +7,9 @@ use std::collections::BTreeMap;
 #[test]
 #[named]
 fn test_borsh_schema_numeric_primitives_signed() -> cargo_near::CliResult {
+    let mut builder = env_logger::Builder::from_env(env_logger::Env::default());
+    let _e = builder.try_init();
+    log::warn!("path {:?}", std::env::args().collect::<Vec<_>>());
```
frol, line 12: Please, refactor as discussed offline

frol (changes requested):

frol (approved):

Issue #185, opened 2024-07-22
Usage metrics


gagdiez:
In `create-near-app` we have an incredibly simple, non-invasive, and anonymous way of gathering usage metrics though `mixpanel`.

Basically, each time `CNA` is used, a small payload is sent to `mixpanel` that includes which type of contract or frontend was created. Source code [here](https://github.com/near/create-near-app/blob/master/src/tracking.ts).

Would be nice to implement the same concept in `cargo near new`, so we can get back statistics from smart contracts created with rust.

frol:
@gagdiez @garikbesson Have anyone flagged privacy concerns regarding the tracking in CLI?

https://github.com/near/near-cli-rs/issues/83 - I remember I started this issue, but then decided to close it as I was not completely comfortable pushing such code to the users. Even an empty HTTP call to mixpanel is already a traceable event which reveals IP, time, and potentially OS.

gagdiez:
@frol never had any privacy concerns

PR #192, opened 2024-08-01
Tracking `cargo near new` usage

garikbesson:
The issue: #185 

How tracking events looks like:
image

UPDATE: sending field `pkg_version` instead of `cli_version`


cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,47 @@
...
+  timestamp: String,
+}
+
+#[derive(Debug, serde::Serialize)]
+struct TrackingData {
+    event: String,
+    properties: MixpanelProperties,
+}
+
+pub(crate) async fn track_usage() {
```
garikbesson, line 20: Do we have to check CI env here?
Like we do in `create-near-app` [here](https://github.com/near/create-near-app/blob/9c482fd0871a206fae04dabc3e91fb775d9b88f9/src/tracking.ts#L14)

Because I didn't find any tests for the `cargo near new` command in workflows

cargo-near/src/commands/new/mod.rs
```diff
@@ -100,6 +102,8 @@ impl NewContext {
             ));
         }
 
+        util::track_usage().await?;
```
garikbesson, line 105: How is it better to call `async` function here?

I added `#[tokio::main]` directive to the `main` function, but there is still a type issue with `interactive_clap::InteractiveClap` because it's a different type and we have to Futures inside the context

cargo-near/src/util/tracking.rs
frol: It won't contribute that much to the total usage of the tool, so I would skip that.

cargo-near/src/commands/new/mod.rs
frol: Let me walk you through my way of thinking:

1. awaiting for the tracker slows down user for no good reason (imagine I am offline or my internet connection is slow)
2. tracker should not fail the new project creation

That said, we need some background task execution here. See how self-update feature is implemented: https://github.com/near/near-cli-rs/blob/fd3ab99b759b69bb41bd29c1dac7edf8199c2ea5/src/main.rs#L78-L81 - it basically starts a new thread, and then in that thread you can create a new async runtime and run the async function there: https://github.com/near/near-cli-rs/blob/fd3ab99b759b69bb41bd29c1dac7edf8199c2ea5/src/commands/tokens/view_near_balance/mod.rs#L22-L24

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,47 @@
...
+    let client = Client::new();
+
+    if let Ok(_) = client
+        .post("https://api.mixpanel.com/track")
+        .json(&tracking_data)
+        .send()
+        .await
+    {
+    } else {
+        println!("Can't send tracking usage event");
```
frol, line 45: Do we want to bother users with that? I would say no

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,47 @@
+use chrono::prelude::*;
```
frol, line 1: Avoid `*` imports

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let client = Client::new();
+
+    if let Ok(_) = client
+        .post("https://api.mixpanel.com/track")
+        .json(&tracking_data)
+        .send()
+        .await
+    {
+    } else {
```
frol, line 44: It makes more sense to have `if let Err` in this case, if you don't plan to handle Ok anyway:

```suggestion
    if let Err(_) = client
        .post("https://api.mixpanel.com/track")
        .json(&tracking_data)
        .send()
        .await {
```

cargo-near/src/commands/new/mod.rs
dj8yfo: @garikbesson  please consider using  `let _detached_thread_handle = ...`, unlike `self-update` feature, 
i.e. avoid joining the `track_usage` thread. 
Reason for this :
1. it will simplify implementation
2.  `cargo near new` now does `cargo update`  as one of the steps , and if that request finishes earlier than  `track_usage`,
it means that most likely `track_usage` endpoint is unreachable

cargo-near/src/util/tracking.rs
dj8yfo: @frol, well, doing a test in CI  to check this specific `track_usage` call result is ok, i.e. that verifying that call was configured properly  and was recorded on some usage tracking remote, sounds like a good idea to me

frol: Yeah, having a test is useful, but that is not what @garikbesson asked and linked in create-near-app

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,47 @@
...
+struct TrackingData {
+    event: String,
+    properties: MixpanelProperties,
+}
+
+pub(crate) async fn track_usage() {
+    let rustc_version = version();
+    let properties = MixpanelProperties {
+      token: "24177ef1ec09ffea5cb6f68909c66a61".to_string(),
+      rustc_version: format!(
```
gagdiez, line 24: For cargo near new we will want to fix the value `contract: "rs".to_string()`

If needed by the `cargo near` team we can track the rust version, if not, we can remove it

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,47 @@
...
+          "{}.{}.{}",
+          rustc_version.major, rustc_version.minor, rustc_version.patch
+      ),
+      os: env::consts::OS.to_string(),
+      timestamp: Utc::now().to_string(),
```
gagdiez, line 29: I assume we can remove the timestamp (except if mixpanel needs it to accept the request)

cargo-near/src/util/tracking.rs
garikbesson: @frol do we need to track the rust version?

cargo-near/src/commands/new/mod.rs
```diff
@@ -100,6 +102,10 @@ impl NewContext {
             ));
         }
 
+        let _detached_thread_handle = std::thread::Builder::new().spawn(|| {
+          util::track_usage()
+        }).unwrap(); //.join();
```
garikbesson, line 107: @frol @dj8yfo 
without joining here it seems like the `track_usage` just isn't invoked

cargo-near/src/util/tracking.rs
frol: @garikbesson I'd track CLI version instead of Rust version.

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn test_tracking() {
+      let thread_handle = std::thread::Builder::new()
+        .spawn(|| track_usage())
+        .unwrap();
+
+      if let Err(e) = thread_handle.join() {
+        panic!("{:?}", e);
+      }
+    }
+}
```
garikbesson, line 86: @dj8yfo Did you mean something like that?

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    #[test]
+    fn test_tracking() {
+        let thread_handle = std::thread::Builder::new().spawn(|| track_usage()).unwrap();
+
+        if let Err(e) = thread_handle.join() {
+            panic!("{:?}", e);
+        }
+    }
+}
```
garikbesson, line 84: @dj8yfo

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,84 @@
...
+    };
+
+    let cli_version: String = match str::from_utf8(&output.stdout) {
+        Ok(v) => v.split(" ").collect::<Vec<_>>()[1].to_string(),
+        Err(_) => "unknown".to_string(),
+    };
+
+    let properties = MixpanelProperties {
+        token: "24177ef1ec09ffea5cb6f68909c66a61".to_string(),
+        cli_version: trim_newline(cli_version),
```
frol, line 52: Wow this is an overkill :-D

Cargo provides build-time env variables, and Cargo.toml version value is among them:

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

cargo-near/src/util/tracking.rs
frol: No need to launch a thread in the test. Just run the function. I think that would be enough

cargo-near/src/util/tracking.rs
garikbesson: And the request was to track CLI version

> I'd track CLI version instead of Rust version.

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,85 @@
...
+            .expect("failed to execute process")
+    } else {
+        Command::new("sh")
+            .arg("-c")
+            .arg("near --version")
+            .output()
+            .expect("failed to execute process")
+    };
+
+    let cli_version: String = match str::from_utf8(&output.stdout) {
```
dj8yfo, line 47: Version of `cargo-near` can be addeded by using `env!("CARGO_PKG_VERSION")`.
It's very strange to try to track `near` version, which even may not be installed and is not required to be installed, to measure the count of `cargo near new` calls.

garikbesson: Thank you for the explanation, I changed it.
I was confused by asking to track CLI version. CLI in my mind is `near-cli-rs` :)

cargo-near/src/util/tracking.rs
```diff
@@ -0,0 +1,58 @@
...
+        properties,
+    };
+
+    let client = Client::new();
+
+    if tokio::runtime::Runtime::new()
+        .unwrap()
+        .block_on(
+            client
+                .post("https://api.mixpanel.com/track")
```
dj8yfo, line 37: the test passes for arbitrary path `https://api.mixpanel.com/tracker/anything/else/arbitrary/path`  too

dj8yfo (dismissed):

cargo-near/src/util/tracking.rs
garikbesson: Yep, also looks like the mixpanel endpoint returns 200 both with the right token and with random one

race-of-sloths:
@garikbesson Thank you for your contribution! Your pull request is now a part of the [Race of Sloths](https://race-of-sloths.com)!
Do you want to apply for monthly streak? Get 8+ score for a single PR this month and receive boost for race-of-sloths!

[<picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://badge.race-of-sloths.com/garikbesson?type=bot&theme=dark">
    <source media="(prefers-color-scheme: light)" srcset="https://badge.race-of-sloths.com/garikbesson?type=bot&theme=light">
    Shows profile picture for the author of the PR
</picture>](https://race-of-sloths.com/profile/garikbesson)


<details><summary>Current status: <i>waiting for scoring</i></summary>

We're waiting for maintainer to score this pull request with `@race-of-sloths score [0,1,2,3,5,8,13]` command. Alternatively, autoscoring [1,2] will be applied for this pull request

</details><details><summary>What is the Race of Sloths</summary>
<p>
Race of Sloths is a friendly competition where you can participate in challenges and compete with other open-source contributors within your normal workflow

For contributors:

- Tag @race-of-sloths inside your pull requests
- Wait for the maintainer to review and score your pull request
- Check out your position in the [Leaderboard](https://race-of-sloths.com/leaderboard)
- Keep weekly and monthly streaks to reach higher positions
- Boast your contributions with a dynamic picture of your [Profile](https://race-of-sloths.com/profile/garikbesson)

For maintainers:
- Score pull requests that participate in the Race of Sloths
- Engage contributors with fair scoring and fast responses so they keep their streaks
- Promote the Race to the point where the Race starts promoting you
- Grow the community of your contributors

Feel free to check [our website](https://race-of-sloths.com) for additional details!
</p>
</details> 

<details>
<summary>Bot commands</summary>

- For contributors
  - **Include a PR:** `@race-of-sloths include` to enter the Race with your PR
- For maintainers:
  - **Invite contributor** `@race-of-sloths invite` to invite the contributor to participate in a race or include it, if it's already a runner.
  - **Assign points:** `@race-of-sloths score [1/2/3/5/8/13]` to award points based on your assessment.
  - **Reject this PR:** `@race-of-sloths exclude` to send this PR back to the drawing board.
  - **Exclude repo:** `@race-of-sloths pause` to stop bot activity in this repo until `@race-of-sloths unpause` command is called

</details>

cargo-near/src/util/tracking.rs
garikbesson: But I tested it and the test itself works. If request fails the test fails as well

cargo-near/Cargo.toml
```diff
@@ -59,6 +59,12 @@ pathdiff = { version = "0.2.1", features = ["camino"]}
 unix_path = "1.0.1"
 url = { version = "2.5.0", features = ["serde"]}
 tmp_env = "0.1.1"
+reqwest = "0.12.5"
+tracing = "0.1.40"
+tokio = { version = "1.0", default-features = false, features = [
+    "rt-multi-thread",
+] }
+tracing-test = "0.2.5"
```
dj8yfo, line 67: can be moved to dev-dependencies

cargo-near/src/util/tracking.rs
frol: @garikbesson Sorry for the confusion, `cargo-near` is also CLI and I meant "this [cargo-near] CLI program version"

frol (approved):
@race-of-sloths score 5

PR #205, opened 2024-08-16
chore: updated near-workspaces-rs

akorchyn:
@race-of-sloths


frol (approved):
@race-of-sloths score 3

PR #285, opened 2025-01-06
chore: unpin `cc` after issue resolution

dj8yfo:


dj8yfo:
resolution was in https://github.com/rust-lang/cc-rs/pull/1336

dj8yfo:
@race-of-sloths include

akorchyn (approved):
Awesome that you kept track of it
@race-of-sloths score 3

PR #289, opened 2025-01-10
fix: remove from env CARGO_ENCODED_RUSTFLAGS for easier nested builds, simplify RUSTFLAGS computation rule

### LINKED ISSUE #287: `error: output of --print=file-names missing when learning about target-specific information from rustc` when building with build.rs
There is something wrong with `cargo-near-build`, it seems. There is a project where it used to work (cargo-near-build there is 0.3.0, but neither it works with that old version nor it works after I upgraded everything cargo-near, cargo-near-build, and the build image):

image

```bash
git clone https://github.com/near-devhub/neardevhub-treasury-dashboard
cd neardevhub-treasury-dashboard
cargo near build reproducible-wasm
```

I tested it on MacOS and Ubuntu, with old cargo-near 0.10.1 + cargo-near-build 0.3.0 and new cargo-near 0.13.2 + cargo-near-build 0.4.2, but it just fails and I cannot figure out what is wrong. Compiling the child contract on its own works just fine.

Here is the branch with the updated cargo-near versions: `refactor/upgraded-reproducible-config`

dj8yfo:
image

resolves  #287 

this stuff gets set for build-scripts https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/custom_build.rs#L378, while `RUSTFLAGS` is unset

---

tested to help resolve problem in #287 on https://github.com/dj8yfo/neardevhub-treasury-dashboard/commit/56bcbcfe3a11e77af16f0b6e6fa4356955b42cd8

### DIFF
--- a/cargo-near-build/src/cargo_native/compile.rs
+++ b/cargo-near-build/src/cargo_native/compile.rs
@@ -25,15 +25,12 @@ where
     let final_env = {
         let mut env: BTreeMap<_, _> = env.into_iter().collect();
         if hide_warnings {
-            env.insert("RUSTFLAGS", "-Awarnings");
+            env.insert(crate::env_keys::RUSTFLAGS, "-Awarnings");
         }
         env
     };
-    // removing env, which may be implicitly passed when bulding from within a build-script
-    // see https://github.com/near/cargo-near/issues/287
-    // CARGO_ENCODED_RUSTFLAGS="-Awarnings" and RUSTFLAGS="-Awarnings" both result
-    // in mysterious failures of `cargo build --target wasm32-unknown-unknown` (**cargo** bug)
-    let removed_env = ["CARGO_ENCODED_RUSTFLAGS"];
+
+    let removed_env = [crate::env_keys::CARGO_ENCODED_RUSTFLAGS];
 
     let artifacts = invoke_cargo(
         "build",
--- a/cargo-near-build/src/cargo_native/compile.rs
+++ b/cargo-near-build/src/cargo_native/compile.rs
@@ -118,6 +118,11 @@ where
     let cargo = std::env::var("CARGO").unwrap_or_else(|_| "cargo".to_string());
     let mut cmd = Command::new(cargo);
 
+    // removing env, which may be implicitly passed when bulding from within a build-script
+    // see https://github.com/near/cargo-near/issues/287
+    // CARGO_ENCODED_RUSTFLAGS="-Awarnings" and RUSTFLAGS="-Awarnings" both result
+    // in mysterious failures of `cargo build --target wasm32-unknown-unknown` (rustc bug)
+    cmd.env_remove("CARGO_ENCODED_RUSTFLAGS");
     cmd.envs(env);
 
     if let Some(path) = working_dir {
--- a/cargo-near-build/src/cargo_native/compile.rs
+++ b/cargo-near-build/src/cargo_native/compile.rs
@@ -15,30 +15,32 @@ use super::ArtifactType;
 pub fn run<T>(
     manifest_path: &ManifestPath,
     args: &[&str],
-    mut env: Vec<(&str, &str)>,
+    env: Vec<(&str, &str)>,
     hide_warnings: bool,
     color: ColorPreference,
 ) -> eyre::Result<CompilationArtifact<T>>
 where
     T: ArtifactType,
 {
-    let mut final_env = BTreeMap::new();
-
-    // this will overwrite any other RUSTFLAGS specified
-    if hide_warnings {
-        env.push(("RUSTFLAGS", "-Awarnings"));
-    }
-
-    // last instance of a key gets inserted
-    for (key, value) in env {
-        final_env.insert(key, value.to_string());
-    }
+    let final_env = {
+        let mut env: BTreeMap<_, _> = env.into_iter().collect();
+        if hide_warnings {
+            env.insert("RUSTFLAGS", "-Awarnings");
+        }
+        env
+    };
+    // removing env, which may be implicitly passed when bulding from within a build-script
+    // see https://github.com/near/cargo-near/issues/287
+    // CARGO_ENCODED_RUSTFLAGS="-Awarnings" and RUSTFLAGS="-Awarnings" both result
+    // in mysterious failures of `cargo build --target wasm32-unknown-unknown` (**cargo** bug)
+    let removed_env = ["CARGO_ENCODED_RUSTFLAGS"];
 
     let artifacts = invoke_cargo(
         "build",
         [&["--message-format=json-render-diagnostics"], args].concat(),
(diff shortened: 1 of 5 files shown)


cargo-near-build/src/cargo_native/compile.rs
```diff
@@ -24,28 +24,14 @@ where
...
-                    if !rustflags.is_empty() {
-                        rustflags.push(' ');
-                    }
-                    rustflags.push_str(value);
-                }
-            }
-            _ => {
-                final_env.insert(key, value.to_string());
-            }
-        }
```
dj8yfo, line 48: this is just a simplification to avoid concatenating with value from env. 
It will be simply `-Awarnings` for abi builds and `-C link-arg=-s` for wasm builds

cargo-near-build/src/cargo_native/compile.rs
```diff
@@ -118,6 +104,11 @@ where
     let cargo = std::env::var("CARGO").unwrap_or_else(|_| "cargo".to_string());
     let mut cmd = Command::new(cargo);
 
+    // removing env, which may be implicitly passed when bulding from within a build-script
+    // see https://github.com/near/cargo-near/issues/287
+    // CARGO_ENCODED_RUSTFLAGS="-Awarnings" and RUSTFLAGS="-Awarnings" both result
+    // in mysterious failures of `cargo build --target wasm32-unknown-unknown` (rustc bug)
+    cmd.env_remove("CARGO_ENCODED_RUSTFLAGS");
```
frol, line 111: The comment says that both `CARGO_ENCODED_RUSTFLAGS` and `RUSTFLAGS` cause problems, but we only remove `CARGO_ENCODED_RUSTFLAGS`, is this intended?

dj8yfo: RUSTFLAGS gets reset to unset in build scripts, so the 2nd or 3rd of 4th build won't inherit this value from 
previous 1st or 2nd or 3rd one, 
and the 1st build isn't having its value concatenated with value from env as per another comment, 
and will have it defined as `-Awarnings` for abi builds and `-C link-arg=-s` for wasm builds

dj8yfo:
tested this on 
https://testnet.nearblocks.io/address/community.devhub-287b.testnet?tab=contract
https://testnet.nearblocks.io/address/uniquehandle.community.devhub-287b.testnet?tab=contract
https://testnet.nearblocks.io/address/discussions.uniquehandle.community.devhub-287b.testnet?tab=contract 
:ok_hand:

dj8yfo:
tested this on 
https://testnet.nearblocks.io/address/community.devhub-287c.testnet?tab=contract
https://testnet.nearblocks.io/address/discussions.uniquehandle.community.devhub-287c.testnet?tab=contract
https://testnet.nearblocks.io/address/uniquehandle.community.devhub-287c.testnet?tab=contract
:ok:

dj8yfo:
somewhat redundantly tested this on https://github.com/dj8yfo/neardevhub-treasury-dashboard/commit/6e8f7657188de46f51340b23756365a48c0f9a69 via build log only (no deploy): 
```bash
     -                Binary: /home/near/code/web4/treasury-web4/target/near/treasury_web4.wasm
     - SHA-256 checksum hex : 94d35d0c174020f04a59cbe18f8408021e3f86a0d426686fb8e9f2a64b0a2323
     - SHA-256 checksum bs58: B1xGfvAYbi4ZktyztiiqGaVbvG7JtXmBuxc3rUV1s7HU
```

```bash
 │ warning: treasury-factory@0.1.0: Build artifact path: /home/near/code/web4/treasury-web4/target/near/treasury_web4.wasm
 │ warning: treasury-factory@0.1.0: Sub-build artifact SHA-256 checksum hex: 94d35d0c174020f04a59cbe18f8408021e3f86a0d426686fb8e9f2a64b0a2323
 │ warning: treasury-factory@0.1.0: Sub-build artifact SHA-256 checksum bs58: B1xGfvAYbi4ZktyztiiqGaVbvG7JtXmBuxc3rUV1s7HU
 │ warning: treasury-factory@0.1.0:
 │ warning: treasury-factory@0.1.0:
 │ warning: treasury-factory@0.1.0: Path to result artifact of build is exported to `BUILD_RS_WEB4_TREASURY_WEB4_WASM`
```

dj8yfo:
@race-of-sloths include

cargo-near-build/src/cargo_native/compile.rs
```diff
@@ -15,44 +15,32 @@ use super::ArtifactType;
...
+    let final_env = {
+        let mut env: BTreeMap<_, _> = env.into_iter().collect();
+        if hide_warnings {
+            env.insert("RUSTFLAGS", "-Awarnings");
         }
-    }
+        env
+    };
+    // removing env, which may be implicitly passed when bulding from within a build-script
+    // see https://github.com/near/cargo-near/issues/287
```
akorchyn, line 33: I think you should add a comment explaining these details. It's not easy to follow just from the issue

cargo-near-build/src/cargo_native/compile.rs
```diff
@@ -15,44 +15,32 @@ use super::ArtifactType;
...
+    // CARGO_ENCODED_RUSTFLAGS="-Awarnings" and RUSTFLAGS="-Awarnings" both result
+    // in mysterious failures of `cargo build --target wasm32-unknown-unknown` (**cargo** bug)
```
akorchyn, line 35: Do we need to create an issue for cargo? Or do we have something to link?

akorchyn: Let's add this info into the README explaining why we remove that and why. Maybe, some people rely on that functionality.  I doubt it, but it's better to have that than not have :)

akorchyn (approved):
@race-of-sloths score 13

Good job!

cargo-near-build/src/cargo_native/compile.rs
dj8yfo: I'll create an issue to create an issue to cargo. These have to be perfect to have a chance of someone taking a look at them, and those take a lot of time to file in

cargo-near-build/src/types/near/docker_build/metadata.rs
```diff
@@ -109,7 +109,7 @@ impl ReproducibleBuild {
         for command_token in build_command {
             if command_token
                 .chars()
-                .any(|c| !c.is_ascii() || c.is_ascii_control() || c.is_ascii_whitespace())
+                .any(|c| !c.is_ascii() || c.is_ascii_control())
```
dj8yfo, line 112: changing this allows doing
```toml
container_build_command = ["cargo", "near", "build", "non-reproducible-wasm", "--locked", "--env", "RUSTFLAGS=--verbose -C link-arg=-s"]
```

PR #290, opened 2025-01-13
ci: fix clippy 1.84

dj8yfo:


dj8yfo:
@race-of-sloths include

race-of-sloths (approved):
@race-of-sloths score 2

PR #350, opened 2025-05-28
feat: halt wrong building

gagdiez:
Currently contracts need to be compiled using `cargo near build`, since using `cargo build` will end up in a built contract that cannot be used

This PR adds a `build.rs` file that forces users to use `cargo near build`, and configures `cargo near` to set a unique `ENVIRONMENTAL VARIABLE` when building

### DIFF
--- a/cargo-near/src/commands/new/mod.rs
+++ b/cargo-near/src/commands/new/mod.rs
@@ -213,6 +213,10 @@ const NEW_PROJECT_FILES: &[NewProjectFile] = &[
         content: include_str!("new-project-template/Cargo.template.toml"),
     },
     NewProjectFile {
+        file_path: "build.rs",
+        content: include_str!("new-project-template/build.rs"),
+    },
+    NewProjectFile {
         file_path: "README.md",
         content: include_str!("new-project-template/README.md"),
     },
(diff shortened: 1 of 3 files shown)


gagdiez:
Here is the main idea, but I need somebody to take over. I created the changes on a branch of this repo so anyone from the project can take over.

The only missing part is: making `cargo near build` create an `env variable` `CARGO_NEAR` which I do not know how to do, and I'm sure somebody with actual knowledge on Rust will be able to tackle in a matter of minutes

dj8yfo:
Looks like building with plain `cargo` could get unblocked in the future https://github.com/near/nearcore/pull/13602 (support of `reference-types` wasm feature), so i'd not try to enforce anything until it's clear if it's supported or not

gagdiez:
we ran a hackathon last week and literally every single team (they were 20
in total) had an issue with this

accordingly to our statistics, more than a hundred contracts are created
per week

I would strongly suggest to either fix it, or enforce it, since it greatly
hurts the dev experience

gagdiez:
@frol has also asked to leave this as last resort, I'll close it now and maybe only implement it in `create-near-app` until a better solution is found

PR #351, opened 2025-06-13
Create rust.yml

Emadalshamery:


frol:
@Emadalshamery We already have https://github.com/near/cargo-near/blob/main/.github/workflows/test.yml, which covers much more than this proposed CI workflow.

Emadalshamery:
Good

PR #357, opened 2025-08-06
feat: Added instructions for users on compiling the project on unsupported versions of Rust.

FroVolod:
Currently, Wasm compiled with version 1.87.0 or later is not compatible with NearCore VM.
Added instructions for downgrading to version 1.86.0 in the contract compilation toolkit.
Added a separate flag  `--skip-rust-version-check` to allow skipping this check.


frol (approved):
@race-of-sloths score 3

PR #368, opened 2025-11-03
use #[derive(Default)] for BuildContext

PolyProgrammist:
fixing the run https://github.com/near/cargo-near/actions/runs/18909632711/job/54264743333?pr=366


frol (approved):
@race-of-sloths score 2

PR #373, opened 2025-11-28
Feat/features abi flag

pityjllk:
Add --abi-features flag to separate ABI and WASM build features

  Summary

  - Add new --abi-features CLI flag to specify features used only during ABI generation
  - When --abi-features is not provided, falls back to --features for backward compatibility
  - Allows building WASM with different feature flags than ABI generation

  Motivation

  Currently, the ABI is built behind an "abi" feature flag that often requires conditional compilation based on #[cfg(not(target_arch = "wasm32"))]. This is because the same feature flags are passed to both
  the ABI generation step (which runs on the host architecture) and the WASM contract build.

  This creates friction when contracts have features that should only appear in the ABI but not in the compiled WASM, or vice versa. For example, a contract might have a gated feature that enables certain
  methods only for testing/development purposes in the ABI documentation, while the production WASM should not include these methods.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4029,9 +4029,9 @@ dependencies = [
 
 [[package]]
 name = "rkyv"
-version = "0.7.45"
+version = "0.7.46"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


frol:
I see that CI failed for Linux. Let's see if macOS tests would complete, if so, I will re-run the Linux CI to see if that was just a glitch.

PR #381, opened 2026-01-09
fix(cargo-near-build): pin `bon` and `cargo-platform`

vsavchyn-dev:
This PR pins `bon` and `cargo-platform` dependencies for compatibility of freshly created template by `cargo near new` with rustc 1.86.0

Original issue was mentioned here: https://t.me/NEAR_Tools_Community_Group/9882

Note: test with `Cargo.template.toml` will fail as the release for `cargo-near-build` has not been done yet.


vsavchyn-dev:
@r-near test will still fail as `cargo-near-build` v0.10 depends on `bon`. I thought it would be better to release template straight up with new release

PR #409, opened 2026-03-23
fix: set CFLAGS to disable bulk-memory for wasm32 C deps

### LINKED ISSUE #408: [BUG] cargo-near should use `-C target-cpu=mvp` and `wasm-opt --signext-lowering` to produce mvp compatible wasm
# cargo-near should use `-C target-cpu=mvp` and `wasm-opt --signext-lowering`

## Summary

`cargo near build` produces WASM binaries containing post-MVP WebAssembly instructions that NEAR's VM rejects with `CompilationError(PrepareError(Deserialization))`. Two categories of unsupported instructions appear in the output:
 
* **Bulk-memory operations** (`memory.fill`, `memory.copy`) 
* **Sign-extension operations** (`i32.extend8_s`, `i32.extend16_s`, `i64.extend8_s`, etc.)

Reference: https://github.com/near/nearcore/issues/8358

**ALSO cargo-near should ideally enforce CFLAGS for dependant crates that are built from c/c++ sources** 

## The problem

Both categories of instructions are emitted by default by LLVM for the `wasm32-unknown-unknown` target since LLVM 16 (Rust 1.70+). The Rust compiler's `-C target-cpu=mvp` flag prevents rustc from emitting them in user code, but Rust's **pre-compiled standard library** (libcore, liballoc, etc.) shipped with every toolchain was built without this flag and already contains these instructions.

### Proof: sign-ext ops in pre-compiled libcore

```bash
$ /opt/homebrew/opt/llvm@18/bin/llvm-ar x \
    ~/.rustup/toolchains/1.85.0-aarch64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libcore-be358420139f7a58.rlib \
    --output /tmp/rlib_check

$ wasm-tools dump /tmp/rlib_check/core-be358420139f7a58.core.17160b5ec18e4cd3-cgu.0.rcgu.o \
    | grep -c "extend8_s"
15
```

15 `i32.extend8_s` instructions are baked into `libcore` at the toolchain level.

### What cargo-near should do

Three changes:

**1. Add `-C target-cpu=mvp` to RUSTFLAGS** 
**2. Set `CFLAGS="-mcpu=mvp"` in the build environment**
**3. Add `--signext-lowering` to the wasm-opt invocation** 

`--signext-lowering` is a dedicated wasm-opt pass that rewrites all sign-extension instructions into equivalent MVP bitwise operations (shift left + arithmetic shift right). Unlike `-O`, this is an explicit lowering pass that guarantees removal of all sign-ext ops, including those from the pre-compiled standard library.

### Effect of both changes

| Source of post-MVP ops | Fixed by `-C target-cpu=mvp` | Fixed by `CFLAGS="-mcpu=mvp"` | Fixed by `--signext-lowering` |
|------------------------|------------------------------|-------------------------------|-------------------------------|
| User Rust code | Yes | — | Yes (redundant) |
| C dependencies via cc crate | No (separate compiler) | Yes | Yes (redundant) |
| Pre-compiled libcore/std | No | — | Yes |

## Environment

- cargo-near: 0.19.0 (cargo-near-build 0.11.1)
- Rust: 1.85.0 / 1.86.0
- Platform: macOS aarch64 (Apple Silicon)
- wasm-opt: 126 (Homebrew binaryen)

r-near:
## Summary

- Set `CFLAGS_wasm32_unknown_unknown="-mno-bulk-memory -mno-bulk-memory-opt"` (and CXXFLAGS) in the build environment to prevent C/C++ dependencies from emitting bulk-memory wasm instructions

## Problem

LLVM 21+ enables `bulk-memory` and `bulk-memory-opt` by default for `wasm32-unknown-unknown`. This causes C dependencies (e.g., the `ring` crate) to emit `memory.fill`/`memory.copy` instructions, which NEAR's VM rejects with `CompilationError(PrepareError(Deserialization))`.

Both flags are needed: LLVM 21 split bulk-memory into two separate features, and `-mno-bulk-memory` alone only disables the first — `memory.fill` still gets emitted through `bulk-memory-opt`. See [llvm/llvm-project#109443](https://github.com/llvm/llvm-project/issues/109443) for related discussion.

Other post-MVP features (sign-ext, mutable-globals) are left enabled as NearVM supports them since protocol 62.

## Testing

Verified with the `register-contract` from [defuse-protocol/near-outlayer](https://github.com/defuse-protocol/near-outlayer/tree/defuse/v0.1.15/register-contract) which depends on `ring` (C crypto code):

| Scenario | Result |
|---|---|
| clang 18 + this fix | Builds and deploys, identical output to without fix |
| clang 21 + this fix | Builds and deploys successfully |
| clang 21 without fix | Fails at wasm-opt with `memory.fill` errors |

Fixes #408


r-near:
nearcore 2.12 supports latest wasmtime. probably not necessary any more

PR #417, opened 2026-05-21
feat: inject `--cfg near` into wasm build by default

### LINKED ISSUE #416: Inject `--cfg near` into wasm build by default
## Background

`near-sdk-rs` PR [near/near-sdk-rs#1534](https://github.com/near/near-sdk-rs/pull/1534) is migrating from a transitive `near-contracts` cargo feature to a `--cfg near` rustc flag for selecting the on-chain host-function code path (hashing via `keccak256`/`sha256`/`ripemd160`/etc. host fns vs the pure-Rust fallback). The flag-based approach removes a feature-cascading burden across downstream crates (see [intents#275](https://github.com/near/intents/pull/275) where Arseny hit 5 transitive features per crate × ~10 crates).

The piece that `cargo-near` needs to do is set `--cfg near` automatically when building contract wasm. Without it, contracts compiled with the new `near-sdk` silently use the pure-Rust hashing path embedded in their wasm — they still produce correct hashes, but every hash burns extra gas and adds binary size.

## The change

In `cargo-near-build/src/near/build/mod.rs` (around line 214), the default RUSTFLAGS for the wasm build step:

```rust
let mut build_env = vec![(env_keys::RUSTFLAGS, "-C link-arg=-s")];
```

becomes

```rust
let mut build_env = vec![(env_keys::RUSTFLAGS, "-C link-arg=-s --cfg near")];
```

Doc comments in `cargo-near-build/src/env_keys.rs` and the README's RUSTFLAGS section (cargo-near/README.md lines 260-269) need matching updates. The integration tests in `integration-tests/` may need adjustments if any of them assert on the default RUSTFLAGS string.

The ABI generation step uses a different RUSTFLAGS (`-Awarnings`, in `cargo_native/compile.rs:28`) and runs against a native target, so it doesn't need `--cfg near`.

## Backwards compatibility

| `cargo-near` | `near-sdk` | Result |
| --- | --- | --- |
| current (no `--cfg near`) | current (uses #1534) | Works but uses pure-Rust hashing in-wasm — bigger binary + extra gas. Same hash output. |
| **new (this fix)** | current (uses #1534) | Ideal. Host functions used. |
| new (this fix) | old (pre-#1534, uses `target_arch = "wasm32"` gates) | Works. Old SDK has no `cfg(near)` references so it ignores the flag. **Caveat:** rustc warns `unexpected_cfgs` unless old SDK declared `check-cfg`. |
| current | old | Status quo. |

The only friction is the third row — new cargo-near + old SDK = `warning: unexpected cfg condition name: near` from rustc. Documented in release notes, addressable by upgrading near-sdk or adding `[lints.rust.unexpected_cfgs] check-cfg = ['cfg(near)']` to the contract crate. Not a hard break.

## Why this can't be done downstream

`cargo-near-build` explicitly does NOT concatenate the default RUSTFLAGS with the ambient environment (documented in `env_keys.rs`):

> RUSTFLAGS aren't concatenated (implicitly) with values from environment

So contracts can't set `--cfg near` via shell env or `.cargo/config.toml` — `cargo-near-build` overrides them. The override-via-`--env`-flag mechanism works but pushes the burden onto every contract author and every test that calls `cargo_near_build::build_with_cli` (e.g. the workaround in [near/near-sdk-rs#1534's commit 129a13a](https://github.com/near/near-sdk-rs/commit/129a13a) which had to inject the env across 6 test files). Centralizing in cargo-near closes that.

## Scope of follow-up

After this lands and ships in a release, the workaround in #1534 can be reverted (the `.cargo/config.toml` rustflag changes and the `build_with_cfg_near` helpers in the workspaces tests).

r-near:
Sets `--cfg near` automatically for the wasm build stage so `near-sdk >= 5.27` selects the on-chain host-function code path. Closes #416 — see that issue for full context.

## What changes

- **Canonical carrier:** `CARGO_ENCODED_RUSTFLAGS` (0x1f-separated, robust against args with spaces) instead of `RUSTFLAGS`.
- **Default tokens:** `-C link-arg=-s` (unchanged).
- **User overrides:** `--env 'RUSTFLAGS=...'` and `--env 'CARGO_ENCODED_RUSTFLAGS=...'` are still honored; ENCODED wins over RUSTFLAGS, matching cargo's own precedence.
- **`--cfg near` is force-appended** to the effective rustflags. It cannot be dropped by a user override.

ABI build stage is unchanged (still uses `RUSTFLAGS="-Awarnings"`).

## Backwards compat

| cargo-near | near-sdk | Result |
| --- | --- | --- |
| new (this PR) | new (≥5.27) | Ideal — host functions used. |
| new | old (pre-#1534) | Works. Old SDK ignores `cfg(near)`. Only friction: `unexpected_cfgs` warning unless the SDK declares `check-cfg`. |
| old | new | Works, but pure-Rust hashing embeds in wasm (bigger binary, more gas). Same hash output. |
| old | old | Status quo. |

## Verified end-to-end

Built `examples/adder` from `near/near-sdk-rs` via the patched binary; per-crate build fingerprints show `rustflags = ["-C", "link-arg=-s", "--cfg", "near"]` — confirms the encoded form decodes correctly and `--cfg near` propagates through every dep.

## Follow-up

After this ships, the workaround in near/near-sdk-rs#1534 (the `.cargo/config.toml` rustflag edits and the `build_with_cfg_near` test helpers) can be reverted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ RUSTFLAGS="your_custom_value" cargo near build non-reproducible-wasm
 ```
 won't result in `"your_custom_value"` affecting the build.
 
-`RUSTFLAGS="-Awarnings"` is always used for abi build stage, and `RUSTFLAGS="-C link-arg=-s"` for wasm build stage.
+`RUSTFLAGS="-Awarnings"` is always used for abi build stage, and `RUSTFLAGS="-C link-arg=-s --cfg near"` for wasm build stage. The `--cfg near` flag is read by `near-sdk` (5.27+) to select the on-chain host-function path; see [near/near-sdk-rs#1534](https://github.com/near/near-sdk-rs/pull/1534).
 
 Logic for concatenating default values of this variable with values from env was removed in `cargo-near-0.13.3`/`cargo-near-build-0.4.3`, as it was seen as
 an unnecessary complication.
@@ -275,7 +275,7 @@ There's still a way to override this parameter for wasm build stage, e.g.:
 
 ```lang
 cargo near build non-reproducible-wasm --env 'RUSTFLAGS=--verbose'
-RUST_LOG=info cargo near build non-reproducible-wasm --env 'RUSTFLAGS=--verbose -C link-arg=-s'
+RUST_LOG=info cargo near build non-reproducible-wasm --env 'RUSTFLAGS=--verbose -C link-arg=-s --cfg near'
 ```
 
 ### `CARGO_ENCODED_RUSTFLAGS`
--- a/README.md
+++ b/README.md
@@ -266,17 +266,25 @@ RUSTFLAGS="your_custom_value" cargo near build non-reproducible-wasm
 ```
 won't result in `"your_custom_value"` affecting the build.
 
-`RUSTFLAGS="-Awarnings"` is always used for abi build stage, and `RUSTFLAGS="-C link-arg=-s --cfg near"` for wasm build stage. The `--cfg near` flag is read by `near-sdk` (5.27+) to select the on-chain host-function path; see [near/near-sdk-rs#1534](https://github.com/near/near-sdk-rs/pull/1534).
+`RUSTFLAGS="-Awarnings"` is always used for abi build stage. For the wasm build stage the
+default is `RUSTFLAGS="-C link-arg=-s"`, plus `--cfg near` is **force-appended** to whatever
+RUSTFLAGS ends up being (the default, or a user override via `--env`). The `--cfg near` flag is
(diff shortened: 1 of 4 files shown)


r-near:
@codex review

vsavchyn-dev (approved):
Tested with defuse contracts and example contracts in [`near-sdk-rs`](https://github.com/near/near-sdk-rs).
No deviations from original build in terms of abi, but there is a small increase of size (~1% in defuse's main contract for instace, so it should be fine).

PR #445, opened 2026-07-09
chore: pin to nearcore 2.13.0 stable

r-near:
nearcore 2.13.0 is on crates.io and the rest of the 2.13 cascade is published (`near-cli-rs` 0.28.0, `near-workspaces` 0.23.0), so this builds against the live stable crates.

- Swaps the RC pins for stable: `near-primitives` `=0.37.0-rc.2` → `0.37.0`, `near-cli-rs` `=0.28.0-rc.1` → `0.28.0`, `near-workspaces` (integration-tests dev-dep) `=0.23.0-rc.1` → `0.23.0`, integration-tests sandbox pin `2.13.0-rc.2` → `2.13.0`.
- Releases `cargo-near` 0.22.0 and `cargo-near-build` 0.11.5 — the latter picks up the `cargo near check` and build-size output work (#442, #441) that merged since rc.1, so it can no longer stay on the published 0.11.4.

### DIFF
--- a/cargo-near/Cargo.toml
+++ b/cargo-near/Cargo.toml
@@ -39,7 +39,6 @@ names = { version = "0.14.0", default-features = false }
 derive_more = "2"
 shell-words = "1.0.0"
 interactive-clap = "0.3.2"
-# nearcore 2.13 (post-quantum) crates on crates.io.
 near-cli-rs = { version = "0.28.0", default-features = false }
 near-primitives = { version = "0.37.0", default-features = false }
 reqwest = "0.12.5"
--- a/cargo-near/Cargo.toml
+++ b/cargo-near/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-near"
-version = "0.22.0-rc.1"
+version = "0.22.0"
 authors = ["Near Inc <hello@nearprotocol.com>"]
 edition.workspace = true
 rust-version = "1.93"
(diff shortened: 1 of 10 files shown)


integration-tests/src/lib.rs
```diff
@@ -14,11 +14,10 @@ pub fn setup_tracing() {
     let _ = cargo_near::setup_tracing(true, false);
 }
 
-/// nearcore version the suite spins its sandbox node on. Pinned to the 2.13 RC
+/// nearcore version the suite spins its sandbox node on. Pinned to 2.13
 /// (protocol v85) so tests exercise cargo-near against 2.13 rather than the
-/// `near-sandbox` crate's 2.12 default (`DEFAULT_NEAR_SANDBOX_VERSION`). Bump to
-/// the stable 2.13 tag once it ships.
-pub const NEAR_SANDBOX_VERSION: &str = "2.13.0-rc.2";
+/// `near-sandbox` crate's 2.12 default (`DEFAULT_NEAR_SANDBOX_VERSION`).
```
vsavchyn-dev:
```suggestion
/// `near-sandbox` crate's 2.13 default (`DEFAULT_NEAR_SANDBOX_VERSION`).
```
Typo?

cargo-near/Cargo.toml
```diff
@@ -39,11 +39,9 @@ names = { version = "0.14.0", default-features = false }
 derive_more = "2"
 shell-words = "1.0.0"
 interactive-clap = "0.3.2"
-# DRAFT(post-quantum): nearcore 2.13 RC crates. Pinned to the published 2.13
-# RCs on crates.io (near-cli-rs 0.28.0-rc.1 pins near-primitives =0.37.0-rc.2).
-# Drop the `-rc` pins for the stable versions once 2.13 ships.
-near-cli-rs = { version = "=0.28.0-rc.1", default-features = false }
-near-primitives = { version = "=0.37.0-rc.2", default-features = false }
+# nearcore 2.13 (post-quantum) crates on crates.io.
```
vsavchyn-dev:
```suggestion
```
Don't think we need to mention that anymore

integration-tests/src/lib.rs
```diff
@@ -14,11 +14,10 @@ pub fn setup_tracing() {
...
+pub const NEAR_SANDBOX_VERSION: &str = "2.13.0";
```
vsavchyn-dev: Tbh, I am a bit against pinning `NEAR_SANDBOX_VERSION` in `cargo-near` and so on - it would be troublesome to later make an easy upgrade to new `near-sandbox`, no?

vsavchyn-dev (changes requested):

integration-tests/src/lib.rs
r-near: yeah I just dropped this pin entirely ffe3ca4

cargo-near/Cargo.toml
r-near: Agreed, removed it in ea09625

integration-tests/src/lib.rs
r-near: That's fair. As long as we update the near-sandbox dep, we should pull in the latest. Made that change here: ffe3ca4

vsavchyn-dev (approved):

PR #453, opened 2026-07-17
fix: keep `CARGO_NEAR_ABI_PATH` stable across runs and skip no-op ABI copies

r-near:
## Problem

Follow-up to #447, which fixed the ABI-generation target-dir collisions but did not stop the full-graph recompilation reported on repos like [near/intents](https://github.com/near/intents) (`make` twice → everything rebuilds).

Root cause: `CARGO_NEAR_ABI_PATH` was set to the embedded-ABI copy inside `--out-dir`. `near-sdk-macros` evaluates `env!("CARGO_NEAR_ABI_PATH")` while compiling **itself** ([abi_embed.rs](https://github.com/near/near-sdk-rs/blob/near-sdk-v5.28.3/near-sdk-macros/src/core_impl/abi/abi_embed.rs#L4-L9)), so rustc records the value in its dep-info:

```
# env-dep:CARGO_NEAR_ABI_PATH=…/target/makenear/tmp.DeThpEM4Mf/defuse_abi.zst
```

near/intents' Makefile passes `--out-dir=$(mktemp -d …)` — a fresh random path on every `make` — so the env value changed every run, dirtying `near-sdk-macros` → `near-sdk` → the entire contract dependency graph, every time.

## Fix

- Point `CARGO_NEAR_ABI_PATH` at the intermediate ABI file under the resolved target directory (`target/near/<pkg>/<pkg>_abi.zst`). That path is identical from run to run, and `abi::write_to_file` already skips rewriting it on identical contents. The copy into `--out-dir` still happens and remains the user-facing artifact/reported path.
- Route `fs::copy` through `copy_to_file` so it inherits the identical-contents skip. Unconditionally rewriting the unchanged `*_abi.zst` bumps its mtime on platforms where `std::fs::copy` doesn't preserve mtime (Linux), which spuriously dirties the contract crate that `include_bytes!`-es it even with a stable out-dir. (Invisible on macOS, where `std::fs::copy` preserves the source mtime.)

## Verification

On near/intents with the defuse contract, using the exact Makefile command with a fresh `mktemp -d` out-dir per run:

| | run 1 | run 2 (new out-dir) |
|---|---|---|
| before | full build | **22 crates recompiled** (near-sdk-macros, near-sdk + downstream) |
| after | full build | **0 crates recompiled** |

Dep-info now records the stable path:
```
# env-dep:CARGO_NEAR_ABI_PATH=…/target/near/defuse/defuse_abi.zst
```

## What this does NOT fix

Alternating between *different* contracts of one workspace (as a full `make` in near/intents does across its 14 contract jobs) still rebuilds `near-sdk-macros` and everything downstream on each switch, because the ABI path legitimately differs per contract while the env var is consumed at `near-sdk-macros`' own compile time (both in its [build.rs](https://github.com/near/near-sdk-rs/blob/near-sdk-v5.28.3/near-sdk-macros/build.rs) via `option_env!` and in `abi_embed.rs` via `env!`). Fixing that requires a near-sdk-rs change: defer the env read into the tokens emitted for the contract crate (e.g. `include_bytes!(env!("CARGO_NEAR_ABI_PATH"))`), so the env-dep lands on each contract crate with its own stable value. This PR is still a prerequisite for that to help: without a stable path, the per-contract env-dep would keep flipping via the random out-dir.

### DIFF
--- a/cargo-near-build/src/fs.rs
+++ b/cargo-near-build/src/fs.rs
@@ -3,10 +3,9 @@ use eyre::WrapErr;
 
 /// Copy a file to a destination.
 ///
-/// Does nothing if the destination is the same as the source to avoid truncating the file,
-/// or if the destination already has identical contents — rewriting an unchanged file would
-/// bump its mtime, which spuriously dirties any compilation unit that `include_bytes!`-es it
-/// (e.g. the embedded ABI on platforms where `std::fs::copy` doesn't preserve mtime).
+/// Does nothing if the destination is the same as the source or already has identical
+/// contents — rewriting would bump the mtime and spuriously dirty anything that
+/// `include_bytes!`-es the file.
 #[cfg(feature = "build_internal")]
 pub fn copy(from: &Utf8Path, out_dir: &Utf8Path) -> eyre::Result<Utf8PathBuf> {
     if !out_dir.is_dir() {
--- a/cargo-near-build/src/fs.rs
+++ b/cargo-near-build/src/fs.rs
@@ -3,18 +3,17 @@ use eyre::WrapErr;
 
 /// Copy a file to a destination.
 ///
-/// Does nothing if the destination is the same as the source to avoid truncating the file.
+/// Does nothing if the destination is the same as the source to avoid truncating the file,
+/// or if the destination already has identical contents — rewriting an unchanged file would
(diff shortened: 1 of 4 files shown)


r-near:
@codex review

